PR #2, opened 2020-02-04
Add clouddriver proto descriptions

maggieneterval:
- straw-descriptions for kubernetes account protos

  Formatting inspired by [the kubernetes proto](https://github.com/kubernetes/api/blob/master/core/v1/generated.proto): each description appears above the element it describes, with newlines between items.

 - err if halconfig is not a file

   Some people (definitely not me) might be very confused for a very long time if they pass an invalid halconfig path and for some reason an error doesn't surface until validation.


ezimanyi (approved):
:tada: Looks great!  Just a few minor inline suggestions, feel free to ignore any/all.

PR #10, opened 2020-03-02
Add copyright header to all files

ezimanyi:
I've followed the recommendation of the CDF regarding copyright headers, and attributed it to "The Spinnaker Authors" without a year.

### DIFF
--- a/api/proto/appengine.proto
+++ b/api/proto/appengine.proto
@@ -1,3 +1,18 @@
+/*
+ * Copyright The Spinnaker Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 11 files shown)


ezimanyi:
As part of preparing this repo for moving under the spinnaker org, I also looked into adding `CONTRIBUTING` and `CODE_OF_CONDUCT` files.  It looks like these are actually set at the org-level in [spinnaker/.github](https://github.com/spinnaker/.github/) which indicates they should [propagate to all sub-repos](https://help.github.com/en/github/building-a-strong-community/creating-a-default-community-health-file).  (I'm not sure exactly how that happens though as I don't see those in all the repos currently under `spinnaker`.)

ezimanyi:
Here is the [copyright guidance](https://github.com/cdfoundation/foundation/blob/master/copyright.md) provided by the CDF.

maggieneterval (approved):
:scroll: :scroll: :arrow_right: :face_with_head_bandage:

PR #13, opened 2020-03-09
feat(protos): add DC/OS, Docker Registry, ECS, Huawei Cloud, and Oracle protos

maggieneterval:
* feat(proto): add DC/OS provider protos 

* feat(proto): add Docker Registry provider protos 

* feat(proto): add ECS provider protos 

* feat(proto): add Huawei Cloud protos 

* feat(proto): add Oracle protos 

* refactor(examples): move halconfig.yml and clouddriver.yml into examples directory 

  Per the [recommended golang project structure](https://github.com/golang-standards/project-layout/tree/master/examples), consolidate example input and output into `examples` directory.

* refactor(proto): rename `persistentstorage` to `persistent_storage` per proto style guide 

  Per Protocol Buffers Style Guide, all proto file names should be
  [lower_case_snake.proto](https://developers.google.com/protocol-buffers/docs/style).

* refactor(proto): specify Provider in top-level provider message names 

  In order to keep each message name as specific as possible, renames each cloud provider top-level provider message to include Provider. For example, `Kubernetes` is now `KubernetesProvider`.


api/proto/docker_registry.proto
```diff
@@ -0,0 +1,94 @@
...
+    string primaryAccount = 3;
+}
+
+// A credential able to authenticate against a set of Docker repositories.
+message DockerRegistryAccount {
+    // The name of the account.
+    string name = 1;
+
+    // (Required) The registry address from which to pull and deploy images
+    // (e.g., `index.docker.io`).
```
ezimanyi: I have this configured as `https://gcr.io` (ie, including the protocol)...does it also work without the protocol, or should we include that in the example here?

api/proto/docker_registry.proto
```diff
@@ -0,0 +1,94 @@
...
+    string address = 2;
+
+    // The number of seconds between polling the Docker registry. Certain
+    // registries are sensitive to over-polling, and larger intervals
+    // (e.g., 10 minutes = 600 seconds) are desirable if you experience rate
+    // limiting. Defaults to `30`.
+    int32 cacheIntervalSeconds = 3;
+
+    // The number of threads on which to cache all provided repositories.
+    // Really only useful if you have a ton of repos. Defaults to 1.
```
ezimanyi, line 34: :balance_scale:

api/proto/docker_registry.proto
```diff
@@ -0,0 +1,94 @@
...
+    // provided, Spinnaker will attempt to read accessible repositories from the
+    // `registries _catalog` endpoint.
+    repeated string repositories = 15;
+
+    // If `true`, Spinnaker will sort tags by creation date. Defaults to
+    // `false`.
+    bool sortTagsByDate = 16;
+
+    // If `true`, Spinnaker will track digest changes. This is not recommended
+    // because it consumes a high QPM, and most registries are flaky.
```
ezimanyi: Is "QPM" here queries per minute?  Maybe this could be replaced by a more generic term (ie, not specific to per minute) such as "because it greatly increases queries to the registry".

api/proto/huaweicloud.proto
```diff
@@ -0,0 +1,157 @@
...
+  string subnetId = 10;
+
+  // (Required) The default security group in which images will be baked.
+  string securityGroup = 11;
+
+  // (Required) The bandwidth size of EIP in which images will be baked.
+  int32 eipBandwidthSize = 12;
+}
+
+// Configuration for a base image for the Google provider's bakery.
```
ezimanyi:
```suggestion
// Configuration for a base image for the Huawei Cloud provider's bakery.
```

ezimanyi (approved):
:100: :tada: :woman_technologist: :1st_place_medal: :1234: :keyboard: :cloud:

api/proto/huaweicloud.proto
maggieneterval: :man_facepalming:

api/proto/docker_registry.proto
maggieneterval: I was also not sure what QPM meant and assumed it was a super common abbrev for "queries per minute" that I had just never heard of...but I like your wording better; updated!

api/proto/docker_registry.proto
maggieneterval: Great callout, [Halyard adds `https://` if neither that nor `http://` is included in the address](https://github.com/spinnaker/halyard/blob/master/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/dockerRegistry/DockerRegistryAccount.java#L51), but for kleat let's make users include the protocol (updated the example here and added to the list of validations/defaults/breaking changes).

maggieneterval:
@ezimanyi thanks for the fast review despite this being a TOME of a PR :books: :weary:

PR #14, opened 2020-03-11
feat(proto): document or remove Halyard-defaulted values

maggieneterval:
I think there are two categories of breaking changes to document when considering a migration from Halyard to Kleat:
1. Changes users will need to make to existing Halyard-generated halconfigs in order to use them as input to Kleat.
2. Changes in the behavior of Halyard vs. Kleat when it comes to defaulting unspecified members of an entity.

I propose that we only document each specific breaking change that falls into category (1) (for now), and instead document a general statement about the differences in defaulting behavior between Halyard and Kleat. This documentation, which I've added to the README, should likely be moved to spinnaker.io when we make Kleat GA.

To that end, let's remove all todos from the code around handling Halyard-defaulted values:
- `AwsProvider. defaultKeyPairTemplate` already has a default in [Clouddriver's config](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-web/config/clouddriver.yml#L80).
- `AwsProvider.defaults.iamRole` now has the same default in [Clouddriver](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-web/config/clouddriver.yml#L67) as Halyard after this [PR](https://github.com/spinnaker/clouddriver/pull/4406) was merged.
- Since the default value for `AwsLifecycleHook. defaultResult` only applies when adding a new lifecycle hook with an unspecified `defaultResult`, I don't think we need to document Kleat's lack of a default here as a breaking change, especially since we are now documenting in the README that Kleat will not supply any default values and we are strongly typing each field.
- Similarly, Halyard defaulted `AzureAccount.regions` to a list of two arbitrary regions, and `AzureAccount. useSshPublicKey` to `true`. For the reasons mentioned in the previous bullet, I do not think we need to document this as a breaking change.


api/proto/aws.proto
```diff
@@ -22,23 +22,14 @@ message AwsProvider {
...
     // List of default regions.
-    repeated AwsRegion defaultRegions = 7;
-
-    // Defaults relevant to the AWS provider.
-    // TODO(mneterval): Can we move to Clouddriver?
-    AwsDefaults defaults = 8;
+    repeated AwsRegion defaultRegions = 6;
 
     // Configuration for AWS-specific features.
-    AwsFeatures features = 9;
```
maggieneterval, line 38: (I stared at this for about five years trying to figure out what looked off, then realized that I skipped `10` the first time around here.)

ezimanyi: Technically this is a breaking change from the proto perspective (re-numbering a field)...but given that we are only ever using this as an intermediary to serialize YAML (and nobody is using this yet) there should be no existing protos out there that this could break so I think this is fine.

Probably once we actually fully release this we'll want to stop re-numbering fields, but I think we should still do so where necessary now.

README.md
```diff
@@ -7,3 +7,29 @@ for an overview of what this tool will do.
...
+- Halyard specified
+`HalConfig.Providers.AzureProvider.AzureAccount.useSshPublicKey` as a `string`
+with `'true'` and `'false'` as acceptable values. Kleat requires this field to
+be specified as a `bool`.
+
+#### Optional Changes
+
+- Halyard provided a default value for every possible member of the halconfig.
+Kleat neither requires nor supplies any default values, which means it is
+acceptable to remove all values for which the behavior of an unspecified
```
ezimanyi: I wonder if something like "for which the microservice's deafult value is acceptable" would be more clear?  Or maybe people  shouldn't need to know exactly what fields affect each microservice?  I just found "behavior of an unspecified field" a bit confusing on first read.

ezimanyi (approved):
📄 🚫 🎉 💯 ☕️

api/proto/aws.proto
maggieneterval: Yes! And if we ever do need to remove fields in the future once we consider this a public API, we should use the [`reserved` keyword](https://developers.google.com/protocol-buffers/docs/proto#reserved) so new fields don't overwrite the number or name of removed fields.

ezimanyi: Oh, nice!  Adding the `reserved` keyword to the list of things I learned today...📝

README.md
maggieneterval: > I wonder if something like "for which the microservice's deafult value is acceptable" would be more clear?

Thanks, that's a lot clearer, updating! And I'm not sure how much people should have to know exactly what fields affect each microservice: in an ideal future world, perhaps the `Halconfig` proto lives in Kleat, but the `Clouddriver` proto and all its Clouddriver-only messages are exported from Clouddriver and live close to the code that will default them.

PR #17, opened 2020-03-17
feat(proto): add artifact protos for Helm, HTTP, Maven, Oracle, S3, and templates

maggieneterval:
* feat(proto): add Helm artifact provider protos 

* feat(proto): add HTTP artifact provider protos 

* feat(proto): add Maven artifact provider protos 

* feat(proto): add Oracle Artifact Provider protos

  Realized while adding the Oracle artifact provider that we would need to disambiguate cloud and artifact providers with the same name, which we will do in the next commit...

* refactor(protos): disambiguate providers from artifact providers

  This is a huge scary diff, but the only actual changes are appending `_provider` to all cloud provider proto files, appending `_artifact_provider` to all artifact provider protos, and adding the word `Artifact` to all artifact provider and account messages.

* feat(proto): add S3 artifact provider protos 

* feat(proto): add artifact template protos


ezimanyi (approved):
🏺 🏺 🏺 💯 🎉  🎡 🖥 ☕️ ☁️ 🏺 🏺 🏺

PR #18, opened 2020-03-27
fix(import): Fix import statement

ezimanyi:
Now that kleat lives under spinnaker it should import from there as well.


ezimanyi:
This broke when I deleted the version that I had under `$GOPATH/src/github.com/ezimanyi`

maggieneterval (approved):
👨 ➡️ ⛵️

PR #19, opened 2020-03-30
refactor(core): Use go modules

ezimanyi:
The main advantage is that we can pin dependencies to explicit versions instead of depending on which version is in the user's go path. (There are likely other things we can do with go modules in the future, but for now the main motivation is just this dependency pinning.)


maggieneterval (approved):
🐻📜🔢💯 🤗 📈

PR #20, opened 2020-03-30
refactor(core): Add package for generated protos

ezimanyi:
protoc-gen-go recommends adding a go_package option to specify the package where output code should be placed. For now, just setting the package to "client" to match the directory, but now that we have a lot of protos, we should probably think about how to better organize this directory.


maggieneterval (approved):
➕📦 🔢 ➡️ 🔡
👨‍💻 💅 💯 💯 💯

maggieneterval:
> we should probably think about how to better organize this directory

+1. Looks like we can use protoc-gen-go's [`--go_opt=paths=source_relative`](https://developers.google.com/protocol-buffers/docs/reference/go-generated#invocation) to preserve whatever directory organization we want to implement for the protos 🎉

PR #21, opened 2020-03-30
feat(core): Use a container to compile protos

ezimanyi:
Compiling the protos requires a number of tools to be installed on the local machine (protoc, protoc-gen-go, protoc-gen-doc). While it's not too difficult to set this up once, differences in version of these tools between different contributors can lead to large diffs in generated output, and also makes is difficult to write tooling that ensures that the generated output is in sync.

This commit creates a Dockerfile containing all the required tools at pinned versions. Now all that is needed to compile the protos reproducibly is to have docker installed.

It also adds a Makefile to make it easy to regenerate the protos. Running 'make proto' will update all generated code based on the current status of the proto files. Running 'make checkproto' will check if the generated code and protos are currently in sync, but will not update any of the generated code. (The idea is that 'make checkproto' can be run as part of pre-merge CI to ensure that these always stay in sync.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 protobuilder:
 	docker build -q -t kleat-protobuilder:latest build/protoc/
 
-proto_command = docker run --rm -v $(abspath api/proto/):/proto -v $(abspath api/client):/output/go -v$(abspath docs):/output/doc kleat-protobuilder:latest
+proto_command = docker run --rm -v $(abspath api/proto/):/proto -v $(abspath api/client):/output/go -v $(abspath docs):/output/doc kleat-protobuilder:latest
 
 proto: protobuilder
 	$(proto_command) update
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,10 @@
+protobuilder:
+	docker build -q -t kleat-protobuilder:latest build/protoc/
+
+proto_command = docker run --rm -v $(abspath api/proto/):/proto -v $(abspath api/client):/output/go -v$(abspath docs):/output/doc kleat-protobuilder:latest
+
+proto: protobuilder
+	$(proto_command) update
+
+checkproto: protobuilder
+	$(proto_command) check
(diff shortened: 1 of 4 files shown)


build/protoc/genproto.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/sh
+
+gen_proto() {
+  output_dir=$1
```
maggieneterval, line 4: Probably really dockerdumb question but I'm having a hard time tracing where we are actually passing the output_dir as an argument to this script...I see we're mounting the /client and /docs directories in the `proto_command` defined in the Makefile but is there somewhere else we specify an argument to this script?

Makefile
```diff
@@ -0,0 +1,10 @@
+protobuilder:
+	docker build -q -t kleat-protobuilder:latest build/protoc/
+
+proto_command = docker run --rm -v $(abspath api/proto/):/proto -v $(abspath api/client):/output/go -v$(abspath docs):/output/doc kleat-protobuilder:latest
```
maggieneterval:
```suggestion
proto_command = docker run --rm -v $(abspath api/proto/):/proto -v $(abspath api/client):/output/go -v $(abspath docs):/output/doc kleat-protobuilder:latest
```

maggieneterval (approved):
⚒ 🐳  🔢 ➡️ 🔤 
👏 👏 👏 👏 👏 👏 👏

build/protoc/genproto.sh
ezimanyi: Good question, this is somewhat confusing bash...I think I've written my yearly quota of bash so I'll need to pick another language if this script gets any more complicated.

We don't pass the `output_dir` to the script itself, this is passed in two places within the script:
* `gen_proto /output` in `update_proto`
* `gen_proto /staging` in `check_proto`

maggieneterval: Oh okay I see, thanks!

PR #22, opened 2020-03-30
chore(core): Update protoc-gen-go to V2

ezimanyi:
Go has just released an updated protobuf API implementation; there are a number of breaking changes that make it incompatible with the prior version (and the import path has changed from github.com/golang/protobuf to google.golang.org/protobuf to indicate this).

Given that we are just starting to implement our protos, and none of the breaking changes affect us, it probably makes sense to upgrade now (before we have any code that does rely on behavior of the old version).

This commit updates us to use the new generator by:
(1) Updating the protoc-gen-go docker container to pull from the new repository.
(2) Regenerating the generated code.
(3) Updating the go modules to account for new imports in this new generated code.


go.mod
```diff
@@ -4,6 +4,7 @@ go 1.14
 
 require (
 	github.com/ghodss/yaml v1.0.0
-	github.com/golang/protobuf v1.3.5
+	github.com/golang/protobuf v1.4.0-rc.4
```
maggieneterval, line 7: Wait, do we now need both `github.com/golang/protobuf` and `google.golang.org/protobuf v1.20.1`?

maggieneterval (approved):
➕ 🗣 ➡️ 🐻

go.mod
ezimanyi: Yeah, that was surprising to me at first, but the [blog post](https://blog.golang.org/protobuf-apiv2) says that:
> google.golang.org/protobuf@v1.20.0 is APIv2. This module depends upon github.com/golang/protobuf@v1.4.0, so any program which uses APIv2 will automatically pick a version of APIv1 which integrates with it.

So I'm pretty sure this is expected if I'm reading that correctly.

maggieneterval: 👍 🤷‍♀

ezimanyi: Ah, here's some more context: https://github.com/protocolbuffers/protobuf-go/commit/baf64d5536c1219775e356cc42c08346524e8e68

maggieneterval: Uh oh, I spy an unassigned TODO in that commit

PR #24, opened 2020-04-03
test(ci): test PR to make sure CI works

maggieneterval:
Just want to make sure that the GH actions actually run; I think they did not run on the prior PR because it included mergify changes ❓


maggieneterval:
Oh never mind, they did run on the last PR, the /actions tab just took forever to populate.

PR #26, opened 2020-04-09
refactor(packages): move non-internal packages to /pkg

maggieneterval:
In addition to the autogenerated Go client, the operator eventually plans to consume Kleat's 
- methods for transforming a halconfig into each service config
- method for validating a halconfig

Let's move those packages from /internal to a new [/pkg](https://github.com/golang-standards/project-layout/tree/master/pkg) directory, which is the standard for library code.


ezimanyi (approved):
🚚 📦 💯 🎈

PR #27, opened 2020-04-21
fix(cmd): Executable should be in package main

ezimanyi:
* fix(cmd): Executable should be in package main

  When re-organizing the directories, we renamed the package of the main binary from main to kleat. While all importable packages need to be in a directory with the same name of the package, the entrypoint of the program must be a function main in a package main.

  the executable anymore as go programs mus

* chore(core): Bump dependencies

  Ran 'go get -u ./...' and committed the change. The protobuf bump did not cause any change in generated code as checked by make proto.

* chore(core): Also bump protoc-gen-go in container

  We should at some point have a better way of keeping these in sync, but since we pulled the new protoc-gen-go, we should also update it in the container that compiles the protos. (Actually it's the reverse, the compiler in the containers is generating code that imports protoc-gen-go. We then need to compile that generated code using a version of protoc-gen-go, which ideally would be the same as the one we used to generate the code.)

  This commit bumps the version in the container, and regenerates the code, the only diff being the comment string indicating the version of protoc-gen-go that was used to compile it.


ezimanyi:
It took me longer than I'd like to admit to figure out why `go install` was a no-op...because there was no longer a top-level executable.  Also bumped the dependencies while I was at it.

cmd/kleat/main.go
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package kleat
```
maggieneterval, line 16: Great catch; sorry/thank you!

maggieneterval (approved):
👩‍💻 🔢 🚫 📦 🐛 
👨‍💻 🔢 ➡️ 📦 🔨 🚫 🐛 
❤️

PR #28, opened 2020-04-22
feat(docs): document poor performance of sortTagsByDate flag

maggieneterval:
As reported in https://github.com/spinnaker/spinnaker/issues/5538, the `sortTagsByDate` flag has prohibitive performance implications for registries with tens or hundreds of thousands of tags. However, we learned from conversations in Slack and at the Kubernetes SIG that users with smaller registries still find this flag valuable. Let's at least add a warning to the documentation, and in the future we can revisit either removing the flag or improving the performance of its implementation if the Docker V2 API ever supports fetching tags chronologically (or even batching requests for individual tags to get their creation date).

Corresponding Halyard PR: https://github.com/spinnaker/halyard/pull/1639


ezimanyi (approved):
🏷️ 🔢 📈 🤯 💯

PR #29, opened 2020-04-23
feat(echo): Add notification config

ezimanyi:
* refactor(core): Use sigs.k8s.io/yaml instead of ghodss/yaml

  There has not been a release of ghodss/yaml in more than 3 years; it looks like kubernetes has started maintaining a fork of it with more recent releases, so let's use that.

  The motivation for this is that I want to use UnmarshalStrict in tests so they catch additional fields that shouldn't be there, but that is not present in the latest release of ghodss/yaml (though is on the master branch there).

* feat(echo): Add functionality to generate echo config

  This commit adds an empty echo config, and adds the functionality for kleat to generate an echo config, and a basic test on that config.

  The code to generate the config (as well as the test) is somewhat duplicative of the code for clouddriver. For now I'm deferring cleaning that up into better reusable components, but will do that before we add a third microservice.

* feat(echo): Add notification config

  This commit adds support for configuring notifications via the halconfig, and having those be written to the echo config.

  As artifacts are just passed-through to echo, and the clouddriver work already strongly-typed this, this commit also passes artifacts through to echo.


api/proto/notifications.proto
```diff
@@ -0,0 +1,50 @@
...
+  // The name of your Slack bot.
+  string botName = 2;
+
+  // Your Slack bot token.
+  string token = 3;
+
+  // Slack endpoint. Optional, can only be set if using a compatible API.
+  string baseUrl = 4;
+
+  // Force usage of incoming webhooks endpoint for slack. Optional, only set if
```
maggieneterval:
```suggestion
  // Force usage of incoming webhooks endpoint for Slack. Optional, only set if
```

api/proto/notifications.proto
```diff
@@ -0,0 +1,50 @@
...
+  // Your Twilio account SID.
+  string account = 2;
+
+  // Your Twilio auth token.
+  string token = 3;
+
+  // The endpoint of the Twilio API. Optional, only set if overriding the default.
+  string baseUrl = 4;
+
+  // The phone number from which the SMS will be sent (i.e. +1234-567-8910).
```
maggieneterval:
```suggestion
  // The phone number from which the SMS will be sent (e.g., +1234-567-8910).
```

api/proto/notifications.proto
```diff
@@ -0,0 +1,50 @@
...
+  string from = 5;
+}
+
+// Configuration for Github status notifications.
+message GithubStatusNotification {
+  // Whether Github status notifications are enabled.
```
maggieneterval:
```suggestion
  // Whether GitHub status notifications are enabled.
```

api/proto/notifications.proto
```diff
@@ -0,0 +1,50 @@
...
+  bool enabled = 1;
+
+  // Your github account token.
```
maggieneterval:
```suggestion
  // Your GitHub account token.
```

maggieneterval (approved):
☎️ 💌 📜 ✅ ✅ ✅ 🎉

PR #30, opened 2020-04-23
feat(echo): Add pubsub config

ezimanyi:
This commit adds pubsub config support to kleat. Currently only Google Pubsub is supported in Halyard, so this commit only adds the same to kleat.


api/proto/pubsub.proto
```diff
@@ -0,0 +1,78 @@
...
+  // the name of the project, and must already be configured.
+  string subscriptionName = 3;
+
+  // The path to a JSON service account that Spinnaker will use as credentials.
+  // This is only needed if Spinnaker is not deployed on a Google Compute Engine
+  // VM, or needs permissions not afforded to the VM it is running on.
+  // See https://cloud.google.com/compute/docs/access/service-accounts for more information.
+  string jsonPath = 4;
+
+  // The acknowledgement deadline as configured on the on the Pub/Sub subscription.
```
maggieneterval:
```suggestion
  // The acknowledgement deadline as configured on the Pub/Sub subscription.
```

maggieneterval (approved):
🍺 ⬇️ 📜 💯

PR #32, opened 2020-04-24
feat(echo): Add GCB account and stats

ezimanyi:
* feat(echo): Add GCB account

  Echo requires GCB accounts to be in its config file, as it listens to their configured Pub/Sub topics to respond to changes in build status.

  This commit adds top-level CI field to the hal config, and implements support for GCB accounts. Support for other CI accounts will be done when we implement the rest of the igor config.

* feat(stats): Add stats to echo config

  For now this only passes through whether stats are enabled or not; we'll need to think more about how to handle generating a unique id for the instance, and passing through the Spinnaker version.


api/proto/gcb_ci_provider.proto
```diff
@@ -0,0 +1,31 @@
+syntax = "proto3";
+package proto;
+
+option go_package = "github.com/spinnaker/kleat/api/client";
+
+import "permissions.proto";
+
```
maggieneterval, line 7:
```suggestion
// Configuration for the Google Cloud Build Provider
```

Not sure how much value the top-level docs add, but just to be consistent with the rest of the protos?

maggieneterval (approved):
🍺 ⬇️ ☁️ 🏗️ 🏦 ✅

api/proto/gcb_ci_provider.proto
ezimanyi: Oh, definitely this is useful, especially since we aren't (at least to date) documenting these fields in the `HalConfig` and `EchoConfig` protos, relying instead on this comment to document.

PR #33, opened 2020-04-24
feat(front50): Add azs and oracle persistent stores

ezimanyi:
This also renames the GCS message to GcsPersistentStore to be more in line with how we've named other things.

It also removes the top-level persistentStore field, and requires each persistent store to be marked as enabled or disabled (with exactly one being enabled) to match what front50 expects. While this is a slight change, hopefully it will be easy to configure as the kleat config will only require the single one that's being used to be in the config (unlike Halyard which will add fields for everything).

I didn't add tests here, as we're about to do some refactoring and potentially will clean up the tests. I'll add tests on this when finishing up the rest of the persistent stores.


ezimanyi:
I'm liberally renumbering proto fields for now because we don't actually have any existing messages to support (and it of course doesn't affect the YAML).  At some point we should stop doing that, but that time is not now.

maggieneterval (approved):
🅰️ ⚡ 🐍 
🔮 🥠 
💯

PR #34, opened 2020-04-24
feat(pkg): export HalToServiceConfigs method

maggieneterval:
- Adds proto for `ServiceConfigs`, which will be a map of service name to corresponding config. Only anticipated client (for now) is the operator.
- Adds `HalToServiceConfigs` method to `pkg` for consumption by the operator (and test).


ezimanyi (approved):
I think this looks great, and I like the idea of including the top-level wrapper as a proto which I would not have though to do but love! 💯

PR #35, opened 2020-04-24
refactor(proto): Organize protos into directories

ezimanyi:
This is a first pass, there are still a couple of places I'd like to reorganize but I'd like to get feedback on this first (particularly given the large diff and potential merge conflicts).

I grouped the proto fiels into logical groups an put them in their own directories, removing rendundant parts of the filenames.

In the case of the top-level configs, I removed Config from the name because go style prefers shorter names and 'config.Front50' is just as clear as (or clearer than) 'config.Front50Config'. We could think about whether to shorten the names of other things as well but I haven't done that in this PR (and they are used less frequently so the diff will be smaller).


maggieneterval (approved):
🔢 🔢 🔢 🤯 
⬇️ 
🔢 📁 💅

PR #37, opened 2020-04-24
refactor(core): Tidy the proto directory

ezimanyi:
Move notifications into their own directory and split then up into separate fields. Also, un-nest some places where we still had nested protos.


maggieneterval (approved):
⛔ 🐦 🏠

PR #38, opened 2020-04-27
refactor(tests): convert parse_hal_test to table-based

maggieneterval:
This is a straw-PR. I'd actually like to suggest we write the inputs/outputs to all these tests as YAML for clarity, and have the inputs to these tests simply be a name, file of halconfig YAML, and file of output-service YAML. Let me know what you think!


ezimanyi (approved):
🪑 ⚾ 💯 🥇

PR #39, opened 2020-04-27
refactor(proto): Clean up name of protos

ezimanyi:
* fix(proto): Update proto package names

  Per the proto style guide, the package of a proto should be the directory in which it lives. As a follow-up to splitting the protos into directories grouping common functionality, update the package in each directory.

* refactor(proto): Clean up name of protos

  Now that the protos are in their own directory/package and are referred to by their package name, we can remove redundant parts of the message name.

  This also helps generate client code that is more in line with go style. For example, a GCS artifact account is now referred to as 'artifact.GcsAccount' instead of 'artifact.GcsArtifactAccount'.

  I adopted a style where the top-level proto configuring something lives in that package with (usually) the same name as the package. For example, artifact.Artifacts, ci.Ci, etc.

  I also changed the message for the Google cloud provider to GoogleComputeEngine to avoid confusion with other Google providers.


maggieneterval (approved):
🧹 💅 📛 🔢 💯

PR #40, opened 2020-04-27
feat(proto): Use protojson to unmarshal protos

ezimanyi:
* feat(proto): Use protojson to unmarshal protos

  Instead of relying on the struct tags that are added by protoc-gen-go, use protojson to marshal/unmarshal objects that implement the proto.Message interface. This has the advantage of correctly marshaling/unmarshaling enums, as well as getting us compatibility with the standard JSON format for proto messages.

  As there is no standard YAML format for proto messages (and YAML is a superset of JSON), there is no protoyaml package. Make a thin wrapper that converts YAML to JSON before using protojson (and the reverse for marshaling). This is effectively what our yaml.Unmarshal was doing before under the hood (to use the JSON struct tags) so the JSON <-> YAML conversion is unaffected, it's jsut hat the JSON <-> Object conversion now uses protojson instead of go's built-in json.

* feat(proto): Add enum types

  Now that we can marshal and unmarshal enum types, replace types that were made strings uniquely to work around the unmarshaling issues.


test/data/halconfig.yml
```diff
@@ -5,7 +5,7 @@ providers:
     enabled: true
     accounts:
     - name: k8s
-      providerVersion: 2
```
maggieneterval, line 8: Not directly related to this change but we can actually remove `providerVersion` since Clouddriver will now default this to `V2`!

api/proto/cloudprovider/kubernetes.proto
```diff
@@ -130,3 +130,8 @@ message KubernetesAccountDockerRegistry {
     // The list of Docker registry namespaces usable as image sources.
     repeated string namespaces = 2;
 }
+
+enum KubernetesProviderVersion {
```
maggieneterval: Should we just delete this, since by the time Kleat is GA v1 will likely be totally unsupported?

maggieneterval (approved):
🔢 ➡️ 🔡 1️⃣ 2️⃣ 3️⃣ ✅ 💯 🎉 

This is amazing!!!!

api/proto/cloudprovider/kubernetes.proto
ezimanyi: Good call, deleted!

test/data/halconfig.yml
ezimanyi: Good call; for now I've left it I guess as a check that we don't fail if it's present, but I removed it from the expected clouddriver output.

PR #42, opened 2020-04-28
refactor(proto): Misc cleanup on protos

ezimanyi:
* fix(proto): Don't delete all generated code if protoc fails

  We currently delete all the generated code before running protoc so that we can start with an empty directory (avoiding cases where we miss deleting an old generated file).

  This means that if protoc fails, we delete all the generated code, which can be annoying. Let's instead generate the code into the staging directory and only move it into the final directory if protoc exits successfully.

* refactor(proto): Move providers to cloudprovider package

* refactor(proto): Make Spinnaker nested under front50

  This message is called 'Spinnaker' but the only thing that unifies what is in it is that it contains the things under 'spinnaker' in the front50 config, which makes it seem reasonable to nest it under front50.

* refactor(proto): Misc cleanup on protos

  Add a few missing comments and alphabetize imports.

  In protos that only contain other non-primitive types, I didn't add a comment on each field. (Ex: Given that the Clouddriver proto has a KubernetesProvider field that is itself commented, it seemed redundant to also add a comment inside the clouddriver proto.)


ezimanyi:
That was close, had a poorly-formatted PR name, good think @spinnakerbot doesn't know about this repo!

maggieneterval (approved):
🔢 💅 💯

PR #44, opened 2020-04-28
refactor(test): consistent package structure for tests

maggieneterval:
Reorganizes package `parse_hal`:
- Splits each `hal_to_{service}` method into its own file.
- Splits each `hal_to_{service}` tests into corresponding file, in `parse_hal_test` package within `parse_hal` directory.
- Potentially controversially, moves "end to end" tests that read from YAML into each `hal_to_{service}_test` file as well, since they are in fact testing the conversion of a hal config to a service config (just from YAML rather than Go structs).
- Moves test YAML into `testdata` file as this directory name seems to be a Go convention (it is ignored when you run `go build`).


ezimanyi (approved):
🧹 💅 📦 🏗️ 💯

ezimanyi:
@Mergifyio refresh

PR #45, opened 2020-04-29
fix(proto): change enabled string to bool

maggieneterval:
fix(proto): change enabled string to bool


ezimanyi (approved):
Whoops...thanks!
🧶 ➡️ 👻 L

PR #47, opened 2020-04-29
feat(front50): Add proto for s3 storage config

ezimanyi:
* feat(front50): Add proto for s3 storage config

  Also fix the text that Halyard will create a bucket for you in the GCS proto; kleat will no longer do this.

* fix(proto): Remove non-ASCII quotes from docs

  Let's avoid any encoding issues.


api/proto/storage/s3.proto
```diff
@@ -0,0 +1,50 @@
+syntax = "proto3";
+package proto.storage;
+
+option go_package = "github.com/spinnaker/kleat/api/client/storage";
+
+// Configuration for an Amazon S3 persistent store.
+message S3 {
+  // Whether this persistent store is enabled.
+  string enabled = 1;
```
maggieneterval:
```suggestion
  bool enabled = 1;
```

ezimanyi:
Oh, I guess I should also add S3 to the function you just added to write out the front50 config, but I'll include that in the PR I'm writing now that tests front50.

api/proto/storage/s3.proto
```diff
@@ -0,0 +1,50 @@
...
+
+  // The name of a storage bucket that your specified account has access to.
+  string bucket = 2;
+
+  // The root folder in the chosen bucket to place all of Spinnaker’s persistent
+  // data in.
+  string rootFolder = 3;
+
+  // This is only required if the bucket you specify doesn’t exist yet. In that
+  // case, the bucket will be created in that region.
```
maggieneterval, line 19: Was it Halyard creating the bucket? (Should we remove the "this is only required..." part?)

maggieneterval (approved):
🐍 🐍 🐍 🔢 🎉 🤗

api/proto/storage/s3.proto
ezimanyi: Wow, I never learn.

maggieneterval:
> Oh, I guess I should also add S3 to the function you just added to write out the front50 config, but I'll include that in the PR I'm writing now that tests front50.

No worries either way; I just wanted to add that function so I could test deploying front50 with a configured persistent store (which is GCS)

ezimanyi:
I just added a commit (right at the second you approved this) to include this in the front50 config because I realized I also forgot to update the perisistent storage and front50 protos which should be in thsi commit.

maggieneterval:
🆗 I looked and I still approve it!

api/proto/storage/s3.proto
ezimanyi: Not, it's actually front50 that's [creating it](https://github.com/spinnaker/front50/blob/d4d33ed34911a5e36fb67d1ee63e5c659a6b3b55/front50-s3/src/main/java/com/netflix/spinnaker/front50/model/S3StorageService.java#L73) so we should probably leave it.

The fact that Halyard was picking a bucket name was confusing and hard to support (as kleat doesn't write back to the config so would have no place to put the generated name, and would make a new one every time).  But this seems reasonable as at least the config stays the same every time.

PR #48, opened 2020-04-29
test(front50): Add unit tests on front50 config generation

ezimanyi:
* test(front50): Add unit tests on front50 config generation

* test(front50): Add end-to-end test of YAML generation for front50

### DIFF
--- a/pkg/parse_hal/hal_to_clouddriver_test.go
+++ b/pkg/parse_hal/hal_to_clouddriver_test.go
@@ -32,9 +32,9 @@ import (
 )
 
 var halToClouddriverTests = []struct {
-	n     string
-	h     *config.Hal
-	wantC *config.Clouddriver
+	n    string
+	h    *config.Hal
+	want *config.Clouddriver
 }{
 	{
 		"Empty hal config",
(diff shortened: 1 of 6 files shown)


pkg/parse_hal/hal_to_clouddriver_test.go
```diff
@@ -148,7 +148,7 @@ func TestHalToClouddriver(t *testing.T) {
 		t.Run(tt.n, func(t *testing.T) {
 			gotC := parse_hal.HalToClouddriver(tt.h)
 			if !reflect.DeepEqual(gotC, tt.wantC) {
-				t.Errorf("Expected hal config to generate %v, got %v", tt.wantC, gotC)
+				t.Errorf("%s: Expected hal config to generate %v, got %v", tt.n, tt.wantC, gotC)
```
maggieneterval: I thought I remembered the error output including the name of the test right above each error (which is why you have to pass some name to `t.Run`, but if this makes it even more readable then I'm on board!

maggieneterval (approved):
❗ 🔙 5️⃣ 0️⃣ 📜 ➡️ 📜 💯

pkg/parse_hal/hal_to_clouddriver_test.go
ezimanyi: Great point! I somehow completely missed that in the output, but now that I see the name is there I agree it makes more sense to leave it out of the actual error.  Thanks for pointing that out, I'm still getting used to what this test output looks like!

pkg/parse_hal/hal_to_clouddriver_test.go
```diff
@@ -148,7 +148,7 @@ func TestHalToClouddriver(t *testing.T) {
 		t.Run(tt.n, func(t *testing.T) {
 			gotC := parse_hal.HalToClouddriver(tt.h)
```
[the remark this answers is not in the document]
ezimanyi: Yup, good point, made the change.

PR #49, opened 2020-04-30
refactor(core): Some tidying of parsing and test logic

ezimanyi:
* style(core): Clean up imports 

  It seems like in some places the imports from within the project are grouped by subdirectory, which doesn't seem to be a style I see anywhere else.

* refactor(core): Remove filesystem access from ParseHalConfig 

  This commit pushes reading files to consumers of ParseHalConfig and lightly cleans up the code in WriteConfigs. (I think some of the change to WriteConfigs reverts something I did recently which was to pass the directory separately to write, but on reflection I think I like how it was before.)

* refactor(core): Remove ParseHalConfig 

  This felt like a pretty thin wrapper that was just another level of indirection for calling code. I think there may be another level of indirection needed somewhere, but for now it might make sense to just push everything to calling code and then see what patterns emerge.

* refactor(test): Clean up variable names and error logging 

  In a few places I think we may have gone overboard with short variable names as I'm often scrolling up to see what a single-letter variable name means. Let's give these names that are a bit more descriptive, but still short.

  Also, the tests all call t.Errorf when something goes wrong. That causes execution to continue, which is good if you are checking an assertion and still want to check later ones, but bad if there's something fundamentally wrong with the test that makes later assertions useless. For example, if we can't read in the echo.yml we should just fail immediately rather than continue and log a failure that the empty string didn't match what was expected. This changes any failure that prevents getting anything meaningful from the test to instead call Fail.  (And also by using .Fail instead of .Failf we can pass the error directly instead of converting it to a string.)

  It's somewhat moot that the last assertion in each test I've left as .Errorf, because the test is over anyway, but it matches the principle above---that the test could continue (ie, checking other conditions) if the assertion failed.

* refactor(core): Make expected YAML test table-based 

  The tests that the YAML output are all doing the same set-up and work; let's make this a table-based test in hal_to_service_configs_test so that we can share that logic. This significantly reduces the amount of boilerplate code for testing each service.


maggieneterval (approved):
🧹 🧪 🔢 ✅ 💯 🎉

PR #50, opened 2020-04-30
feat(orca): Add orca protocol buffers

ezimanyi:
This commit adds support for configuring orca with kleat.


api/proto/config/orca.proto
```diff
@@ -0,0 +1,31 @@
...
+  Defaults default = 3;
+
+  // Configuration for pipeline templates.
+  message PipelineTemplates {
+    // Whether pipeline templates are enabled.
+    bool enabled = 1;
+  }
+
+  // Defaults applicable to the orca microservice.
+  message Defaults {
```
ezimanyi, line 21: Let me know what you think about the nested protocol buffers here.  I know at one point we un-nested because the style guide said that messages should rarely be nested, but in this case the messages are really wrappers that are just defining how orca expects the data so it seemed reasonable to nest?  But I'm definitely interested in your opinion here, I can also see an argument for un-nesting these as `OrcaDefaults` and `OrcaBakeDefaults`.

pkg/parse_hal/hal_to_orca.go
```diff
@@ -0,0 +1,46 @@
...
+		Default:           getDefaults(h),
+		PipelineTemplates: getPipelineTemplates(h),
+		Webhook:           h.GetWebhook(),
+	}
+}
+
+func getDefaults(h *config.Hal) *config.Orca_Defaults {
+	if !h.GetProviders().GetAws().GetEnabled() {
+		return nil
+	}
```
ezimanyi, line 31: I'm not sure if it's necessary to return `nil` if AWS is not enabled; I don't think there's any harm in setting this field anyway if AWS is not enabled from a functional perspective.  I did this mostly to avoid generating config that was not needed (similarly to how we set the pipeline template struct to `nil` if it's not set).  This really only affects humans reading the output of `kleat` so if we think we're adding complexity without equal benefit we can remove it and just always set the defaults to this return value below (and similar for pipeline templates).

api/proto/config/orca.proto
maggieneterval: Nesting seems right for these  👍

api/proto/features.proto
```diff
@@ -0,0 +1,21 @@
...
+message Features {
+  // Enable pipeline template support. Read more at https://github.com/spinnaker/dcd-spec.
+  bool pipelineTemplates = 1;
+
+  // Fields below this line are not yet used in any service that kleat supports
+  // TODO: Ensure that they are actually needed and move above this line
+
+  bool auth = 2;
+  bool fiat = 3;
+  bool chaos = 4;
```
maggieneterval, line 16: "Enabling chaos" makes me laugh

pkg/parse_hal/hal_to_orca.go
maggieneterval: I like it the way it is; I think the complexity is worth it 🤷

maggieneterval (approved):
➕ 🐳 🔢 🤗

PR #51, opened 2020-05-01
feat(security): Add security to hal config

ezimanyi:
This commit adds the security block to the hal config, as well as the configuration nested under it. This includes SSL for gate/deck, as well as authentication and authorization.

The one exception is that I left a TODO for OAuth as Halyard looks to be doing some complex logic there and I want to look at separately.


ezimanyi:
I got a bit carried away and ended up doing most of the security protos as that's the next big thing we'll need to do before we can do fiat, gate, or deck.

api/proto/security/authn.proto
```diff
@@ -0,0 +1,183 @@
...
+  // The password used to access the file specified in -keystore.
+  string keyStorePassword = 6;
+
+  // The name of the alias under which this server's private key is stored in
+  // the -keystore file.
+  string keyStoreAliasName = 7;
+
+  // The address of the Gate server that will be accesible by the SAML identity
+  // provider. This should be the full URL, including port, e.g. https://gate.org.com:8084/.
+  // If deployed behind a load balancer, this would be the laod balancer's address.
```
maggieneterval:
```suggestion
  // If deployed behind a load balancer, this would be the load balancer's address.
```

api/proto/security/authn.proto
```diff
@@ -0,0 +1,183 @@
...
+  // The password for the LDAP manager user.
+  string managerPassword = 7;
+
+  // The part of the directory tree under which group searches should be performed.
+  string groupSearchBase = 8;
+}
+
+// Configuration for authentication via X509 certificates.
+// X509 authenticates users via client certificate and a corresponding private key.
+// These certificates optionally provide authorization information via custom Oids
```
maggieneterval:
```suggestion
// These certificates optionally provide authorization information via custom OIDs
```

maggieneterval (approved):
➕ 🔒 💂 ⛵ 📜 💯 

Wow, you got so much done yesterday! 🎉

PR #52, opened 2020-05-01
feat(security): Add OAuth2 protos

ezimanyi:
* fix(proto): Fix package for webhook

  It's in the security directory, so should be in package security.

* feat(security): Add OAuth2 protos

  This commit adds the protos required for configuring OAuth2 through kleat. As I noted in a prior PR, Halyard does some complex logic here that we'll need to account for.

  What it does is that, based on your OAuth2 provider, it automatically sets a bunch of these fields.  So for example if you choose Google as the OAuthProvider, it sets the accessTokenUrl to https://www.googleapis.com/oauth2/v4/token for you.

  I think the way we should handle this is to move all that defaulting to Gate. If Gate finds that the OAuth2 provider is Google, and you have not explicitly set an accessTokenUrl it should set it to the one that Halyard would have set.  (And similarly for the other providers and fields.)

  It's only really for the custom provider that you'll need to set all of these fields. The next step after this PR will be to make these changes in Gate. After I do that, I'll come back and update the comments to explain when you do/don't need to set these fields, but need to make the change in Gate first so I understand this enough to document it.


api/proto/security/authn.proto
```diff
@@ -29,8 +29,90 @@ message OAuth2 {
...
   bool enabled = 1;
 
-  // TODO: Fill in the rest of this, as it looks like there is some reasonably
-  // complex logic in Halyard around this.
+  // Configuration for your OAuth2 client.
+  OAuth2Client client = 2;
+
+  // The map of requirements the userInfo request must have. This is used to
+  // restrict user login to specific domains or having a specific attribute.
+  map<string, string> userInfoRequirements = 3;
```
maggieneterval, line 37: a new specimen!!

maggieneterval (approved):
😮 🔐 🔢 ✅ 💯

PR #53, opened 2020-05-04
fix(oauth2): Update some documentation of the OAuth 2.0 provider

ezimanyi:
Improve the documentation with things that I learned while looking
into how to improve the spinnaker.io documentation around OAuth 2.0.
In particular:
* Add better comments to a few fields
* Convert clientAuthenticationScheme to an enum as there are only
four possible values
* Standardize on OAuth 2.0 in documentation


api/proto/security/authn.proto
```diff
@@ -196,6 +216,7 @@ message Ldap {
   // find the user using user-search-filter, if set.
   string userDnPattern = 3;
 
+
```
maggieneterval: accidental newline?

maggieneterval (approved):
I will greatly simplify thing for both users and contributors.

api/proto/security/authn.proto
ezimanyi: indeed





fixed!

PR #54, opened 2020-05-06
chore(core): Bump dependencies

ezimanyi:
Bump the go packages as well as the version of protoc-gen-go we install in the docker container.


maggieneterval (approved):
🗻 🚸 ⬆️ 💯

PR #55, opened 2020-05-06
 feat(gate): Add initial gate config

ezimanyi:
* fix(gate): Update SAML proto

  Halyard did some mappings and transformations that we won't be replicating in kleat. Notably, Halyard parsed the metadata string you passed on the command line to store it in either metadataLocal or metadataRemote; but then when generating the gate config it would just add whichever one was not null to metadataUrl. We'll only support metadataUrl in kleat and users will need to (once) manually update their key from metadataLocal/metadataRemote to metadataUrl.

  Halyard accepted a redirect URI for SAML and parsed out the port, host, and protocol. We'll require that users supply these separately, which is how Gate expects them.

* feat(gate): Add initial gate config

  This commit adds the ability to configure Gate using kleat. I still need to add tests for gate, as well as do some more extensive comparison with what Halyard outputs to make sure everything is correct here, but this defines all of the types and allows a gate config to be generated.


maggieneterval (approved):
🤺 🔢 🔐 ✅ 💯 💯 💯 
Wow, these were definitely the most complicated set of protos yet!

PR #56, opened 2020-05-08
feat(proto): add CI protos

maggieneterval:
Reunites the GCB protos with long-lost friends:
- Codebuild
- Concourse
- GCB
- Jenkins
- Travis
- Wercker


api/proto/ci/travis.proto
```diff
@@ -0,0 +1,41 @@
...
+  // (Required) The base URL to the Travis UI (https://travis-ci.org).
+  string baseUrl = 3;
+
+  // The GitHub token with which to authenticate against Travis.
+  string githubToken = 4;
+
+  // (Required) The number of repositories the Travis integration should fetch
+  // from the API each time the poller runs. Should be set a bit higher than the
+  // expected maximum number of repositories built within the poll interval.
+  int32 numberOfRepositories = 5;
```
maggieneterval, line 34: Another one where 0 vs. unset matters, which I've dealt with (for now) by marking it as required.

api/proto/ci/gcb.proto
```diff
@@ -30,7 +28,4 @@ message GoogleCloudBuildAccount {
   // Google Compute Engine VM, or needs permissions not afforded to the VM
   // it is running on.
   string jsonKey = 4;
-
-  // Fiat permissions configuration.
-  Permissions permissions = 5;
```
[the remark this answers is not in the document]
ezimanyi: While there aren't any commands, you can still set it by directly editing the halconfig YAML, so it might make sense to keep this here to avoid breaking users who have done that.

api/proto/ci/jenkins.proto
```diff
@@ -0,0 +1,39 @@
...
+  // The name of the account.
+  string name = 1;
+
+  // (Required) The username of the Jenkins user to authenticate as.
+  string username = 2;
+
+  // (Required) The password of the Jenkins user to authenticate as.
+  string password = 3;
+
+  // (Required) The address at which the Jenkins master is reachable.
```
ezimanyi: Should we call this a Jenkins server?  (And maybe in the config field too, though I guess that would break existing configs.)

ezimanyi (approved):
🌊 👁️ 🥇 💯

api/proto/ci/gcb.proto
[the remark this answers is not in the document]
maggieneterval: 🆗 ✅

api/proto/ci/jenkins.proto
maggieneterval: So the only reason I left this one as `master` but changed all the other references to that word is because the Jenkins docs are the only docs that actually use that word to refer to a Jenkins instance, but you're right and server is just as good so I changed it.

I didn't change the field name (here or on the other CI protos) because it seems like it's a slippery slope from there to renaming all the other poorly-named fields...

ezimanyi: Yeah, good point on renaming the fields 👍

PR #57, opened 2020-05-12
feat(proto): Add protobuf-dev to proto-building container

ezimanyi:
This package is needed to be able to use well-known protobuf types, as these are not included in the protoc package, which only has the compiler itself.


maggieneterval (approved):
🔢 💪

PR #58, opened 2020-05-12
fix(proto): Wrap some proto values

ezimanyi:
This commit wraps some proto fields in cases where the service needs to make a meaningful distinction between the default value for the type and no value for the type (ie null).


maggieneterval (approved):
🤡 ➖ F ➕ 🅱️ ✅
🆗 ✅ 
❗ 🆗 ✅ 
❓ 🆗 ✅

PR #60, opened 2020-05-13
refactor(proto): Allow protoc version to be specified

ezimanyi:
* refactor(proto): Allow protoc version to be specified

  We currently can't specify the version of protoc we use to compile the protcol buffers; we specify a version of alpine, then get whatever version of protoc comes with that version of alpine.

  Because alpine doesn't have a lot of the same libraries as on other linxu distributions, it generally requires specially-compiled binaries which is why we need to install the one from apk add rather than just download the latest protoc release.

  To get a specific version, we could check out the protoc source and build it on alpine; I tried this when I was first setting this up and while it works it's really slow. (Of course that image would be cached but I'd rather not have 'make proto' take ~10 mins even occasionally, which is why I did apk add.)

  I think the real fix here is just to use debian instead of alpine, so we can just use the standard protoc releases. This has the additional advantage that we can also just pull the desired releases of protoc-gen-go and protoc-gen-doc from github instead of building them. (Before we *had* to build them from source in the absence of an alpine-compatible release; we still could but it's much faster to just grab the tarball from github and extract it.)

  Finally, note that we are still using alpine as the builder container to fetch and unzip the releases. That's fine because we're not using it to build anything, so we can use the files we fetched in that container on the Debian container. (This would not be the case if we actually built protoc-gen-go or protoc-gen-doc from source; we'd then need to use a debian machine to do that, or figure out some way to cross-compile.)

* chore(proto): Bump protoc

  Bump protoc to 3.11.4 now that we can explicitly bump it. This adds no changes, other than the comment.


maggieneterval (approved):
👟 💬 ✌️ 🔢 ✅

maggieneterval:
> I think the real fix here is just to use debian instead of alpine, so we can just use the standard protoc releases. This has the additional advantage that we can also just pull the desired releases of protoc-gen-go and protoc-gen-doc from github instead of building them.

💯 thanks for the explanation!

PR #61, opened 2020-05-14
test(gate): Add test for gate config generation

ezimanyi:
* fix(proto): Fix some of the security protos 

  Halyard is writing the edited enums in upper case, so we should make our enum the same. (I was confused because the toString was writing it in lower-case, but I guess Halyard is not using toString as part of serialization, which I guess probably makes sense when you think about it enough.)

  I also accidentally named a property spring instead of security. Let's also rename the message SpringSecurity and un-nest it as it represents the top-level security.* config settings for Spring (to which I'll be shortly adding another field).

* feat(proto): Add basic authentication 

  Halyard currently looks on the service settings file for a flag to determine whether to enable basic authentication; this is different from all the other authentication methods, which are configured together in the same block of the hal config.

  As I don't anticipate wanting to expose the service settings through kleat (as these are in general related to how a service is deployed rather than how it behaves) let's add a new field to the halconfig so users can still configure it if they want.

  I suspect this is not very common in any real production set-up
  (as it would require sharing a username and password with everyone using Spinnaker) but as it was easy to just add the two fields, it seemed better than not supporting it anymore. Users *will* need to move any basic auth config from the service settings file to the hal config.

* feat(gate): Write out the gate config to file 

  Also, update the logic for generating the gate config to avoid generating config blocks that are not relevant. In general, I think we should adopt the pattern (and sort of already have) where if we're not just looking something up in the hal config and are actually creating a new struct inline, we should first check to see if we need to create the wrapper struct to avoid writing out a bunch of empty structs.

  Remove the TODO as upon further investigation we don't need to allow these to be settable in kleat and can rely on the default port and binding.

* test(gate): Add test for gate config generation 

  This commit adds the end-to-end test that the gate config is correctly generated when everything is specified. (Of course this isn't really realisitic as you likely won't have every type of auth configured but given that kleat just passes things through it's still a good test.)

  Next will be the unit tests that verify the particular parts of the config generation that are somewhat complex work properly.

* fix(tests): Use more specific comparators in tests 

  Protobufs support a proto.Equal method that uses proto-supported reflection to deeply compare two messages; use this instead of reflect.DeepEqual as go reflection should not be used on protobufs in favor of protobuf-specific reflection.

  Similarly, in one case we're comparing bytes with reflect.DeepEqual whereas it would be better to just use the specific comparator for bytes. (I remember thinking this when I wrote it but I guess I never changed it.)


pkg/parse_hal/hal_to_service_configs_test.go
```diff
@@ -86,6 +86,10 @@ func TestHalToServiceYAML(t *testing.T) {
 			"orca.yml",
 			services.GetOrca(),
 		},
+		{
```
maggieneterval, line 89: It's awesome that this was the only non-yaml needed to add these tests, I'm getting super on board with these table tests

pkg/parse_hal/hal_to_clouddriver_test.go
```diff
@@ -16,13 +16,13 @@
 package parse_hal_test
 
 import (
-	"reflect"
```
maggieneterval, line 19: Yaaay bye reflection, great find on the better comparators 🎉

api/proto/security/authn.proto
```diff
@@ -289,3 +292,20 @@ message Iap {
   // The URL containing the Cloud IAP public keys in JWK format.
   string iapVerifyKeyUrl = 5;
 }
+
+// Configuration for basic username/password authentication
+message Basic {
```
maggieneterval, line 297: > Users *will* need to
move any basic auth config from the service settings file to the
hal config.

Should we add this sentence to the list of breaking changes in the README?

maggieneterval (approved):
🧪 🤺 🔢 🏗️ ✅ 🎉 💯 🤗

api/proto/security/authn.proto
ezimanyi: Yes! Definitely, and I also need to move the few other things in our Google doc to the README.   I probably should have done it in this PR, but I'll do it in the next PR that adds unit tests.

PR #62, opened 2020-05-14
feat(proto): add canary protos 

maggieneterval:
* feat(proto): add canary protos 

* docs(config): update README with canary migration instructions 

  The existing canary section of the halconfig is difficult to represent in proto. Let's require users to migrate each integration block they care about from an item in a list to an entry in a map.


api/proto/config/kayenta.proto
```diff
@@ -0,0 +1,41 @@
...
+    canary.Stackdriver stackdriver = 2;
+    canary.Gcs gcs = 3;
+    canary.Prometheus prometheus = 4;
+    canary.Datadog datadog = 5;
+    Aws aws = 6;
+    canary.S3 s3 = 7;
+    canary.SignalFx signalfx = 8;
+    canary.NewRelic newrelic = 9;
+
+    message Google {
```
maggieneterval, line 31: So, this is a little annoying and I'm open to other ways of structuring the messages here. A halconfig canary.Google block is different than a Kayenta canary.Google block because we only need `enabled` and `accounts` to populate the Kayenta block (and derive the Stackdriver and GCS blocks from the other top-level fields). Similar story with AWS and S3. This will probably be clearer once I write the translation logic for hal --> kayenta, so let me know if this would be easier to review if I added that as a third commit to this PR.

api/proto/canary/newrelic.proto
```diff
@@ -0,0 +1,35 @@
...
+  // (Required) Your account's unique New Relic Insights API key. See
+  // https://docs.newrelic.com/docs/insights/insights-api/get-data/query-insights-event-data-api.
+  string apiKey = 3;
+
+  // (Required) Your New Relic account ID. See
+  // https://docs.newrelic.com/docs/accounts/install-new-relic/account-setup/account-id.
+  string applicationKey = 4;
+
+  // Configuration for the New Relic Insights server endpoint.
+  message Endpoint {
```
ezimanyi, line 32: I wonder if this could become a top-level message as this is the second account I've seen it in to date.

api/proto/canary/gcs.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
+package proto.canary;
+
+option go_package = "github.com/spinnaker/kleat/api/client/canary";
+
+// Configuration for the GCS canary integration. If enabled, you must
+// also configure at least one canary.GoogleAccount with a list of
+// supportedTypes that includes canary.SupportedType.CONFIGURATION_STORE and/or
+// canary.SupportedType.OBJECT_STORE.
```
maggieneterval, line 9: This is another confusing thing worth calling out: Halyard doesn't actually expose commands to edit any accounts' `SupportedType`s. It defaults it for metrics-only integrations, which I handled by defaulting those types in Kayenta itself. When you enable GCS, S3, or Stackdriver using Halyard, it automatically adds the appropriate types to each configured account's list. The reason I exposed the field on `google` and `aws` is because I think there is a reasonable use case for using different accounts for `CONFIGURATION_STORE` vs. `OBJECT_STORE` (Halyard just adds both to any account when a persistent store is enabled at the top-level). This won't impact existing Halyard users since they will already have these specified in their halconfig, but I've added explicit documentation for GCS, S3, and Stackdriver so this is clear for new users who will need to fill in supported types manually.

api/proto/config/kayenta.proto
ezimanyi: I think this makes sense, and is similar to what I did for authentication, where there is a `google` block under `Gate` that would never really mean anything else outside of Gate so I made it a nested message there.

ezimanyi (approved):
🐦 💯 🥳 ✅ 🎈 🏅

api/proto/canary/gcs.proto
ezimanyi: I think this makes sense! While it may take a few minutes for a user to understand this the first time they set it up, I think it's reasonable to expect that level of understanding and will probably be helpful to those users in the long run.

api/proto/canary/newrelic.proto
maggieneterval: I thought about that too. I think when we consider sharing messages we need to consider the most common cases of if/how these configuration parameters will be updated. I think sharing an Endpoint message would make sense if we anticipate a future where someone refactors all the client code at once that uses the current structure. That would be a nice future. But I think the future we'll get is one-off changes from provider maintainers who will only update the calling code for their particular integration, and won't even have a way to test other integrations.

maggieneterval:
Sorry, forgot to rebase off your protoc version bump, which explains why `make checkproto` was passing locally but failing on the CI server. Should be good to go now!

api/proto/canary/newrelic.proto
ezimanyi: That's a fair point, I guess for now it's only on field under this but perhaps in the future there might be integration-specific fields.

PR #64, opened 2020-05-14
fix(docs): Update protoc-gen-doc

ezimanyi:
This upgrade is not just because there's a new version, but actually fixes something that has been (slightly) annoying me whenever I look at the docs markdown. This latest version removes a lot of the extraneous whitespace so it's easier to read the docs as markdown.


maggieneterval (approved):
🚫 🏳️ 👩‍🚀 ❤️

PR #65, opened 2020-05-15
feat(fiat): Generate fiat config

ezimanyi:
* refactor(security): Split authentication and authorization

  The security package is fairly large; let's split it into authentication and authorization.

  One motivation for this is that you can only use a single enum constant once in each package (not counting nested enums), so the likelihood of collision goes up as packages get larger.

* refactor(authz): Nest RoleProviderType enum

  This enum is very specific to the message in which it's used, so let's nest it.

* feat(fiat): Add fiat config proto

  This commit just adds the proto that holds the fiat config; future commits will actually add the logic to generate a fiat config.

* feat(fiat): Generate fiat config

  This commit adds the logic for generating a fiat config from the hal config, and adds a basic-end-to-end test.

* test(fiat): Add unit tests to fiat config generation

  As there is no complex logic in the config generation a simple test of an empty config and a non-empty config should be sufficient.


maggieneterval (approved):
💵 📜 ✅ 💯 🤗

PR #66, opened 2020-05-19
feat(pkg): translate and write kayenta.yml from halconfig 

maggieneterval:
* feat(proto): add kayenta to services proto 

* feat(pkg): translate and write kayenta.yml from halconfig 

* feat(test): add end-to-end test for kayenta config translation 

* feat(test): add table tests for kayenta config


pkg/parse_hal/hal_to_kayenta.go
```diff
@@ -0,0 +1,90 @@
...
+
+package parse_hal
+
+import (
+	"github.com/spinnaker/kleat/api/client/canary"
+	"github.com/spinnaker/kleat/api/client/config"
+)
+
+func HalToKayenta(h *config.Hal) *config.Kayenta {
+	if !h.GetCanary().GetEnabled() {
```
maggieneterval, line 25: I'm not sure about this yet. An alternative to returning an empty Kayenta config when canary is disabled is to return nil here. I'm not sure how much it matters. Either way, we'll want to decouple kleat's config translation logic from however we decide to handle having users conditionally deploy the Kayenta microservice. So in this new paradigm, toggling canary on or off will not cause Kayenta to not be deployed at all like it is with Halyard, since we don't want the kustomize config to know anything about the output of kleat. That said, we still need to support the top level canary enabled flag in Kleat because it's used in Deck.

maggieneterval: Another alternative is we make write_configs a little smarter, and don't even write a kayenta.yml if canary is disabled. I don't want to make a final decision until I figure out the kustomize workflow but please let me know if you have thoughts.

ezimanyi: I think this is reasonable for now.  I think there are a few places we'll need to think about how to control what gets deployed, but we should think of them as a whole; I don't really have any great thoughts about it yet.

ezimanyi (approved):
🔡 ➡️ 🔠 📜 💯 🥳

PR #67, opened 2020-05-21
feat(pkg): include optional service enabled as config property

maggieneterval:
We have (for now) decided to have whether or not an optional service is enabled be a property of any services that need that information, rather than service discovery information in spinnaker.yml. Let's translate canary.enabled into services.kayenta.enabled for the two services that need to communicate with Kayenta (Gate and Orca).


api/proto/config/service_enabled.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto3";
+package proto.config;
+
+option go_package = "github.com/spinnaker/kleat/api/client/config";
+
+// Configuration for whether a particular microservice is enabled.
+message ServiceEnabled {
```
ezimanyi, line 7: Do we anticipate that this could ever have other information?  Or maybe you deliberately chose to include `Enabled` in the name to make it clear that this is not a place for service-discovery type information?

ezimanyi (approved):
💯 🥳 🥇 🎈

api/proto/config/service_enabled.proto
maggieneterval: Yeah, I wish there was a way to lock this proto so no other fields end up in here...definitely should only include `enabled`; if we feel like other stuff belongs in here I think we need to re-evaluate the entire approach of this even being part of individual service settings.

ezimanyi: Yup, makes sense! Probably overkill to add `reserved 2 to 536,870,911` to the message 😁 .

PR #68, opened 2020-05-21
feat(gate): Add deck base URL

ezimanyi:
* refactor(proto): Add baseUrl field to ServiceEnabled and rename

  We will in some cases need to have kleat set the baseUrl field for a service (in cases where the hal config affects this, such as gate and deck). As this lives under the same config key as whether the service is enabled, let's augment that proto to supply both these fields. Also rename it as it's more general than when it was initially designed.

* feat(gate): Add deck base URL

  Gate needs to know the base URL for deck so that it can handle redirects during auth. Rather than force the user to manually override this in spinnaker.yml, let's output this as part of the gate config.


testdata/gate.yml
```diff
@@ -64,5 +64,7 @@ google:
     audience: projects/0/apps/0
     iapVerifyKeyUrl: https://my-url.test/
 services:
+  deck:
+    baseUrl: https://spinnaker.test:9000
```
maggieneterval, line 68: My only question was going to be, where's the halconfig.yml update?! But just checked and you already did that in a prior PR! 🎉 🎉

maggieneterval (approved):
🚢 🔗 ✅

PR #69, opened 2020-05-26
feat(rosco): Generate rosco config

ezimanyi:
* feat(rosco): Add rosco config protos

  The rosco config is very similar to the clouddriver config, in that it contains the configured cloud providers at the top level. The one difference is that only some cloud providers support baking, so it only contains these providers.

* feat(rosco): Generate rosco config

  This commit adds support for generating a rosco config from a hal config.

* feat(rosco): Add rosco output to service configs

  Also add rosco.yml to the end-to-end test.


ezimanyi:
The main complexity here was going to be handling packer templates, but we are already copying those to the container at build time [here](https://github.com/spinnaker/rosco/blob/5c6d37e26d1627bf1a580ed62ef20e20d9a2b8ce/Dockerfile.slim#L5), so we don't need to worry about building/versioning the templates that ship with Spinnaker.

maggieneterval (approved):
🐶 📸 📜 ✅ 💯

PR #71, opened 2020-05-27
chore(protoc): Bump protoc and protoc-gen-go

ezimanyi:
This gets us to 3.12.x of protoc, which adds support for optional fields. We said we'd revisit in a few weeks (a few weeks ago) and make a final decision about how to handle this. By pulling in a version of protoc that has support for optional fields (behind a flag) it will be easier to play around with this and see the generated code than if we also had to bump protoc.

Also bump protoc-gen-go while we're generating this code.

As usual, neither dependency bump changes the generated code, other than the versioned comment.


maggieneterval (approved):
🔢 ⬆️ 🎉 🙏

PR #72, opened 2020-06-02
fix(cmd): add validation for command line arguments

micnncim:
There's no validation of command line arguments, so panic occurs if the arguments are wrong.

### DIFF
--- a/cmd/kleat/main.go
+++ b/cmd/kleat/main.go
@@ -16,12 +16,18 @@
 package main
 
 import (
+	"fmt"
 	"os"
 
 	"github.com/spinnaker/kleat/internal/write"
 )
 
 func main() {
+	if len(os.Args) != 2 {
+		fmt.Fprintf(os.Stderr, "arguments must be <halPath> <dir>\n")
+		os.Exit(1)
+	}
+
 	hal := os.Args[1]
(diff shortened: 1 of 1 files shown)


maggieneterval (approved):
Thanks so much for the contribution, @micnncim! (And congrats on making the first non-Google contribution to this project!)

Once the initial MVP of Kleat is complete, we had planned to perhaps use the [Cobra](https://github.com/spf13/cobra) library for Kleat's CLI, which has its own format for command line argument validation, so we will likely replace this validation at some point in the medium-term. That said, I am happy to merge this as-is for now since it is certainly an improvement, but will defer to @ezimanyi for final approval/merge.

Thanks again and we're so excited that you're interested in Kleat! 🎉

micnncim:
@maggieneterval Thank you for your review and reply! I'm also so excited to hear your welcome message 😄 

I use Spinnaker in my company and I'm familiar with spf13/cobra (and sure Go), so I'm willing to more contribute to this project if you don't mind.

Thanks for the awesome project 👍

ezimanyi (approved):
Thanks for this!  Improving the CLI code beyond this really simple function is definitely a good idea, but even until we do that this is a great improvement.

PR #74, opened 2020-06-03
feat(pkg): Add Deck support (part 2!)

maggieneterval:
* fix(cmd): fix error check 

  os.Args actually includes the program name as argument 0, so we actually need to check if len != 3.

* fix(proto): add mptV2UI feature flag to Deck proto 

* test(pkg): add Deck end to end tests 

  Although the final settings.js output will not be YAML, let's add an end to end test for the actual data translation from halconfig to Deck config as yaml-to-yaml, like the other microservices. Also adds a few more TODOs to hal_to_deck that should be addressed in a followup PR but do not want to block on completing the writing of settings.js

* feat(internal): write Deck config as JS 

  Kleat can now write Deck's config as a valid JS file.


maggieneterval:
Coming soon in part 3: todos resolved, hal_to_deck tests added! Sorry for the huge cliffhanger!

ezimanyi (approved):
🧗 ❗ 💯 🥳

PR #76, opened 2020-06-05
feat(pkg): add Deck support (part 3!)

maggieneterval:
* feat(pkg): conditionally set Gate basUrl to http or https 

* feat(pkg): add hal-to-deck tests 

* refator(proto): do not expose canary.stagesEnabled 

  Upon reflection, I highly doubt any Halyard users are using Kayenta separately from using Spinnaker's built-in canary stages. Let's not expose this parameter until presented with evidence for its necessity.

* refactor(proto): remove completed TODOs 

  We have already documented the new /var/secrets convention, and the fact that provider defaults statically provided to Deck will fall back to those of the explicitly specified primaryAccount.

* feat(pkg): pass top-level Spinnaker version to Deck 

  Deck displays the top-level Spinnaker version in the UI. Pass top-level Spinnaker version from Halconfig to Deck.


ezimanyi (approved):
♠️ ❤️ ♦️ ♣️ 🎴 💯 🥳

PR #77, opened 2020-06-05
refactor(pkg): remove roscoMode

maggieneterval:
When https://github.com/spinnaker/deck/pull/8333 is merged, Deck will enable Rosco UI support by default.


ezimanyi (approved):
💯 🥳 🦖

PR #78, opened 2020-06-05
feat(deck): Generate environment variables for deck

ezimanyi:
In addition to requiring client-side config provided as settings.js, deck requires some settings to be applied server-side. These are applied by passing environment variables to the deck server process.

There does not appear to be a way to pass these as YAML, we we'll need to generate a file of key=value pairs, which will then be used by a configMapGenerator.

Rather than manually translate the field names to the UPPER_SNAKE case needed for the variables, I've just explicitly overridden the json name for these fields. (Naming the fields in UPPER_SNAKE didn't work as the underscores were removed in the auto-generated JSON name.)


maggieneterval (approved):
♠️ 🔤 ➡️ 📦 ✅ 🤗

PR #79, opened 2020-06-08
fix(deck): Use override URL for gate if specified

ezimanyi:
If an override URL for gate is specified, it should be used in the deck config.

Also, I updated the default URL to be localhost; as this is used to set a client-side URL it will be accessed from outside the cluster where the cluster-local address likely won't be routable. Setting this to localhost allows you to port forward to the cluster to access gate.

I'd imagine that anyone actually running a real instance has the override set (rather than forwarding) but I think localhost is probably a better default (and is what Halyard does).

### DIFF
--- a/internal/write/write_configs.go
+++ b/internal/write/write_configs.go
@@ -67,6 +67,9 @@ func WriteConfigs(halPath string, dir string) error {
 	if err := write(services.GetRosco(), filepath.Join(dir, "rosco.yml")); err != nil {
 		return err
 	}
+	if err := write(services.GetDeck(), filepath.Join(dir, "deck.yml")); err != nil {
+		return err
+	}
 	if err := writeDeck(services.GetDeck(), filepath.Join(dir, "settings.js")); err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


internal/write/write_configs.go
```diff
@@ -67,6 +67,9 @@ func WriteConfigs(halPath string, dir string) error {
 	if err := write(services.GetRosco(), filepath.Join(dir, "rosco.yml")); err != nil {
 		return err
 	}
+	if err := write(services.GetDeck(), filepath.Join(dir, "deck.yml")); err != nil {
```
maggieneterval: Wait, do we need this?

maggieneterval (approved):
🌈  🏍️ 💻 🔗 💯 🎉 🤗 
Thank you for fixing this, and sorry for completely missing it!

internal/write/write_configs.go
ezimanyi: Whoops...we do not!  I added this to help with debugging a test and forgot to remove it.  Thanks for catching it!

PR #80, opened 2020-06-08
fix(deck): Features should be singular

ezimanyi:
The deck config expects feature flags to be under the singular 'feature'. (This is a bit confusing because Halyard used plural 'features' in its config, but it ended up singular in the generated config.)


maggieneterval (approved):
1️⃣ 🎉

PR #82, opened 2020-06-09
feat(pkg): add Igor support

maggieneterval:
* feat(proto): add Repository and Artifactory protos 

  Add repository and artifactory protos in preparation for adding Igor config. Although Halyard allows you to configure Fiat permissions on each artifactory search, this does not appear to be consumed in any way in Igor, so I've omitted this field from Kleat.

* feat(proto): add Igor service config 

  Only includes artifact templates, unlike Halyard, which includes all configured artifact types, because templates are all that appear to be needed by Igor.

* feat(pkg): translate and write Igor config 

* test(pkg): add unit and end to end tests for Igor 

  Also updates the Echo end-to-end tests, since Echo needs the ci.gcb config.


ezimanyi (approved):
💯 🤫 🥳 🥇

PR #83, opened 2020-06-09
refactor(config): Clean up some of the config generation

ezimanyi:
* feat(proto): Add proto for serialized config files

  In order to make it easier to pass around the serialized form of the config files (as well as the name of each file), make a proto to hold this information.

* refactor(config): Add Serializer interface

  There are currently three different ways to serialize a config to bytes:
  * Convert to YAML
  * Write out JavaScript that loads JSON
  * Write out an env file

  Create an interface Serializer and write three implementations of it to abstract away these three different ways of serializing.

* refactor(config): Generate serialized configs as a proto

  Add a function GenerateConfigFiles that converts a proto of the typed service configs into a proto of the serialized service configs. Then use this function when generating the service configs from the hal config.

* refactor(config): Split hal parsing into its own function

  Let's expose a function to parse the hal config from a file and then require that we pass a parsed hal config to the function that writes out the service configs.

* refactor(config): Remove validate_paths

  There are only two functions in this module; one is unused after some recent refactoring and the other can be moved to the one place where it is used.


maggieneterval (approved):
💅 🔢 ➡️ 🔢 💯
Love it!

PR #84, opened 2020-06-09
refactor(core): Move config generation to core

ezimanyi:
* refactor(core): Move config generation to core

  Until we find a valid use case for clients needing to individually generate service configs, let's move the code to generate the individual configs to the internal/ folder, and only expose the top-level functions for generating all configs.

  Let's also shorten the name of all of the hal_to_* files.

* refactor(core): Rename parse_hal and validate_hal

  Partly this is because go discourages package names including underscores
  (excluding _test packages). In the case of parse_hal, rename it transform as it's primarily used to transform protos.

* refactor(core): Add godoc and fix lint errors

  This commit adds godoc to all packages in internal/. As part of this it adds a newline before the copyright and the package name as I learned that otherwise the copyright statement is interpreted as a package-level comment.

  Most of the lint errors in this directory were just missing comments, but fixed a couple of other things (in particular capitalizing URL).

  Also rename a couple of packages to be less verbose.

### DIFF
--- a/internal/serializer/serializer.go
+++ b/internal/serializer/serializer.go
@@ -45,8 +45,8 @@ func (yaml) Serialize(m proto.Message) ([]byte, error) {
 
 type settingsJs struct{}
 
-// SettingsJs is a Serializer that serializes a proto.Message by outputting a JavaScript file to be consumed by Deck
-// .
+// SettingsJs is a Serializer that serializes a proto.Message by outputting a
+// JavaScript file to be consumed by Deck.
 var SettingsJs Serializer = new(settingsJs)
 
 func (settingsJs) Serialize(m proto.Message) ([]byte, error) {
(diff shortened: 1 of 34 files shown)


internal/serializer/serializer.go
```diff
@@ -39,6 +45,8 @@ func (yaml) Serialize(m proto.Message) ([]byte, error) {
 
 type settingsJs struct{}
 
+// SettingsJs is a Serializer that serializes a proto.Message by outputting a
```
maggieneterval, line 48:
```suggestion
// SettingsJs is a Serializer that serializes a proto.Message by outputting a JavaScript file to be consumed by Deck
```
?

maggieneterval (approved):
💬 💯 🎉 📛 ➡️ 📛

internal/serializer/serializer.go
ezimanyi: Ha, good catch!  Thanks!

PR #85, opened 2020-06-09
test(core): Improve debuggability of unit tests

ezimanyi:
* test(core): Create named type for test cases

  All of the config generation tests define the same struct of test cases inline; let's just give it a name and re-use it across the tests to reduce duplication.

  Also remove halTo... from the variables holding the test cases for brevity.

* test(core): Add configTest type

  This type wraps a config generator and a slice of tests that test that generator.

  Update each of the tests to specify the generator and the actual tests. As this changes the indentation of all the tests, this commit will look much cleaner ignoring whitespace.

* test(core): Pull common test logic to a function

  All the config generator tests do the exact same thing to a configTest stuct; pull this logic into a helper function so it's not repeated in every test file.

* test(core): Improve debuggability of unit tests

  If we find that two config protos are not equal, print a diff of them instead of the full contents of both protos.

### DIFF
--- a/internal/convert/clouddriver_test.go
+++ b/internal/convert/clouddriver_test.go
@@ -26,11 +26,7 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
-var halToClouddriverTests = []struct {
-	name string
-	hal  *config.Hal
-	want *config.Clouddriver
-}{
+var clouddriverTests = []testCase{
 	{
 		"Empty hal config",
 		&config.Hal{},
@@ -139,7 +135,7 @@ var halToClouddriverTests = []struct {
 }
 
 func TestHalToClouddriver(t *testing.T) {
-	for _, tt := range halToClouddriverTests {
+	for _, tt := range clouddriverTests {
 		t.Run(tt.name, func(t *testing.T) {
 			got := convert.HalToClouddriver(tt.hal)
 			if !proto.Equal(got, tt.want) {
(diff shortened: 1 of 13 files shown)


ezimanyi:
Let me know if I got too carried away refactoring away the common logic here...technically didn't need to do that to improve the debuggability  😃

internal/convert/clouddriver_test.go
```diff
@@ -26,38 +26,50 @@ import (
...
-		"Empty hal config",
-		&config.Hal{},
-		&config.Clouddriver{},
-	},
-	{
-		"Empty providers",
-		&config.Hal{
-			Providers: &cloudprovider.Providers{},
+var clouddriverTests = configTest{
+	generator: func(h *config.Hal) proto.Message { return convert.HalToClouddriver(h) },
```
ezimanyi, line 30: This is a place where I miss generics.  What my Java brain really wants is for `generator` to be of type:
```
<T extends proto.Message> func(h *config.Hal) T
```
If that were the type, then I could just pass `convert.HalToClouddriver` directly; but since I can't write that I need to write a new function of signature `func(h *config.Hal) proto.Message` that just calls `convert.HalToClouddriver` (and relies on implicitly converting to the interface in the return).

maggieneterval (approved):
🚫 🐛 📏 🧪 👍 
I guess the two costs to this refactor are:
1. When attempting to debug some unit test, a developer must now click through two files to fully understand what the test is doing, whereas before each test's input types and execution logic were confined to its own file.
2. As you mentioned, the expected return type of each test is now implicit rather than explicit.

That said, after reading through this twice, I think consolidating the previously repeated testing logic is likely worth those costs. 👍 💯

PR #86, opened 2020-06-10
style(core): Fix remaining lint errors

ezimanyi:
* refactor(core): Move validate package to internal

  As we deferred any anything except our proof-of-concept validation, this is definitely something we should keep as an implementation detail until we've thought more about it. (We potentially should even delete it as it's a bit strange that we're only validating one field, but at least let's make it internal.)

  Also fix the lint failure, which doesn't like repetition in package/function names (validate.ValidateHalConfig).

  Finally change some names to lowercase so they aren't exported as they aren't used outside the package.

* style(core): Add documentation to fix lint errors

  Add documentation to exported members, fixing the last lint errors in kleat.

* style(core): Fix inspections

  I ran the IntelliJ inspections, and it found three issues, which I fixed since I'm on a roll with minor cleanup. They are:
  * Explicitly assign the result of fmt.Printf to _ to indicate that we are ignoring the error (as there's not much point handling it right before os.Exit anyway). Also use Printf instead of Fprintf as we aren't formatting
  * Rename 'diff' variable which collides with package name
  * Change unused 'h' to _ in function signature (though obviously this will need to be updated when the TODO is handled).

### DIFF
--- a/cmd/kleat/main.go
+++ b/cmd/kleat/main.go
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package main
 
 import (
--- a/cmd/kleat/main.go
+++ b/cmd/kleat/main.go
@@ -25,7 +25,7 @@ import (
 
 func main() {
 	if len(os.Args) != 3 {
-		fmt.Fprintf(os.Stderr, "arguments must be <halPath> <dir>\n")
+		_, _ = fmt.Fprint(os.Stderr, "arguments must be <halPath> <dir>\n")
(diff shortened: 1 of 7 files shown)


maggieneterval (approved):
> I ran the IntelliJ inspections, and it found three issues

I can look into this more when I configure linting/formatting in CI, but perhaps there's some way to have those autocleanups take care of the type of issues found with IntelliJ inspections? Doesn't seem great for the long-term maintenance of the project if we expect Kleat to adhere to those IntelliJ-specific inspection rules and a maintainer needs to periodically make a cleanup commit like this.

PR #87, opened 2020-06-10
feat(ci): run linting in CI

maggieneterval:
Golangci-lint appears to be a popular choice for running gofmt and [various go linters](https://golangci-lint.run/usage/linters/) in CI with a single config file (see list of [companies](https://golangci-lint.run/product/trusted-by/) using it). Let's give it a shot and see if the annotations it adds to PRs are helpful. Alternatively, instead of using this GitHub action, we can add a `make` command as we did for `checkproto` to run the various linters using this config file instead, but thought we could see if the out-of-the-box annotations justify use of this GitHub action.


ezimanyi (approved):
💯 🥇 
I think this is great!

PR #88, opened 2020-06-11
feat(release): add release automation configuration

maggieneterval:
* feat(release): add initial configuration for release with GoReleaser and GitHub Actions. 

  On semver-compliant tag push, the GoReleaser GitHub Action will trigger a GoReleaser release. We can update the .goreleaser.yml config to target a non-default matrix of OS and architecture if necessary in the future.

* feat(docs): update README to state Kleat's readiness for use in dev clusters!

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,25 @@
+name: Kleat Release
+
+on:
+  push:
+    tags:
+      - v*
+
+jobs:
+  release:
+    if: startsWith(github.repository, 'spinnaker/')
+    runs-on: ubuntu-latest
+    steps:
+        uses: actions/checkout@v2
+      - name: Unshallow
+        run: git fetch --prune --unshallow
+      - name: Install Go
+        uses: actions/setup-go@v1
+        with:
+          go-version: 1.14
+      - name: Create release
+        uses: goreleaser/goreleaser-action@v2
+          with:
+            args: release --rm-dist
+          env:
+            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,6 +1,2 @@
-before:
-  hooks:
-    - go mod tidy
-
 builds:
 - main: ./cmd/kleat/main.go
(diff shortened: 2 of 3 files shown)


maggieneterval:
I did a dry-run locally using `goreleaser --snapshot --skip-publish --rm-dist` and a /dist of binaries and other artifacts were successfully created, so we should hopefully be all set to trigger our first Kleat release when the `0.1.0` tag is pushed! 🥳

.github/workflows/release.yml
```diff
@@ -0,0 +1,25 @@
...
+    tags:
+      - v*
+
+jobs:
+  release:
+    if: startsWith(github.repository, 'spinnaker/')
+    runs-on: ubuntu-latest
+    steps:
+        uses: actions/checkout@v2
+      - name: Unshallow
```
ezimanyi, line 14: I think this is equivalent to adding the following to the prior step (to get all history):
```
with:
  fetch-depth: 0
```
But then again, looks like this is what `go-releaser` has in their docs, so might be better to stay with what you have.  (Mostly just looked at this because I wasn't familiar with the `--unshallow` flag of git, which I learned today.)

ezimanyi (approved):
🎣 ➡️ 🐟 🌊 💯 🥳

.github/workflows/release.yml
maggieneterval: I also learned that today! I did end up going with `--unshallow` per `goreleaser` docs, but I see that `fetch-depth: 0` is what `actions/checkout` has in their docs, so no strong opinion 🤷

.goreleaser.yml
```diff
@@ -0,0 +1,6 @@
+before:
+  hooks:
+    - go mod tidy
```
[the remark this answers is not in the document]
maggieneterval: Very glad you asked this because I only decided to include this hook because the GitHub CLI also includes this hook, but looking more closely I think I misunderstood the behavior: I thought this would error out if the modules were untidy, but in fact it just fixes them, so that is definitely not what we want...also, pretty sure we would have caught any untidiness in our now very robust CI, so in any case I will remove this ↪️ 🎣 s.

.github/workflows/release.yml
ezimanyi: No strong opinion either, fine leaving this as is! 💯

PR #90, opened 2020-06-11
feat(cmd): add Cobra framework

maggieneterval:
* feat(cmd): add Cobra framework 

  This commit introduces the Cobra CLI framework, and reorganizes /cmd to adhere to Cobra-recommended standards. Moves primary functionality to the generate subcommand, and introduces the version subcommand, which will print the version/commit/date as injected by goreleaser at build time.

* refactor(release): remove non-default main config 

  Now that we have a root main.go per Cobra standards, we no longer require a non-default path to our main.go to be specified as the build entrypoint.


ezimanyi (approved):
🐍 🖼️ 🧑‍🏭 🧑‍💻 💯 🥳 💻

PR #91, opened 2020-06-11
feat(proto): Pull version of protoc-gen-doc allowing proto3 optional

ezimanyi:
* chore(proto): Bump protoc

* style(proto): Break long lines in shell script

  This will make it easier to see what changed in the next commit
  (and any future changes we make here).

* feat(proto): Add experimental flag

  As we've decided to use the new proto3 optional semantics to handle nullable fields, we need to pass this flag until the feature is fully GA.

* feat(proto): Compile protoc-gen-doc instead of using release

  We'll need to temporarily use a fork of protoc-gen-doc so that we can include optional fields in proto3. As I don't want to actually make official releases of the fork, we'll want to build in the container from a specific tag/commit.

  This is what we used to do, before I changed it to pull the release
  (which is faster). This will be a bit slower the first time you run 'make proto' but should be temporary. Once the changes are in a released version upstream, we can go back to pulling an official release.

  If for some reason these changes don't end up upstream and it becomes an issue, I can make a release in my fork, but want to avoid that assuming this is temporary.

* feat(proto): Pull version of protoc-gen-doc allowing proto3 optional

  This commit pulls in a version of protoc-gen-doc that allows proto3 optional protos. I've created a tag on my fork v1.3.2-proto3optional.1 corresponding to the changes I've submitted upstream. (The idea being that any other changes would be come v1.3.2-proto3optional.2, and if I rebase off of v1.3.3 it would become v1.3.3-proto3optional.1 but hopefully this will get merged quickly upstream so we won't need to update this much).


maggieneterval (approved):
🔢 ❓ ➡️ 📝 👨‍💻 🍴

PR #93, opened 2020-06-12
feat(docs): specify Halyard functionality not available in Kleat 

maggieneterval:
* refactor(docs): move migration instructions to separate README 

  Let's create a separate migrating/README for instructions for migrating from Halyard so as not to clutter the primary README.

* feat(docs): specify Halyard functionality not available in Kleat 

  To help explain to prospective users the narrower problem space of Kleat, explicitly list the Halyard functionality that is being replaced by kustomize, kubectl, and Spinnaker documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,84 +13,8 @@ channel in [Spinnaker slack](join.spinnaker.io).
 
 ### Migrating from Halyard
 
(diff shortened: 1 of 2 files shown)


ezimanyi (approved):
📜 💯

PR #94, opened 2020-06-12
test(core): Improve tests on config generation

ezimanyi:
* fix(deck): Deck environment should have extension .env 

  It's confusing that this is called deck-env.yml as it's not a valid YAML file. Change the name to deck.env.

* test(core): Use ParseHalConfig in tests 

  Instead of manually parsing the hal config, use the ParseHalConfig function.

* test(core): Add an integration test on config generation 

  We have two tests currently that test config generation:
  * A unit test that ensures that the contents of config.Services is the same as individually calling each of the service config generators.
  * An integration test that ensures the serialized YAML for each service matches the expected result in testdata.

  Let's add a test intermediate between these that does the following:
  * Takes test halconfig
  * Generates the service configs as protos
  * Ensures these protos match those obtained by deserialzing the data in the test directory.

  Note that we're not actually testing serialization here, just using YAML as a convenient format for describing the expected output of the integration test. (This is why we read deck-env.yml and deck.yml as YAML files even though the actual output of kleat here is not YAML.)

* test(core): Update config tests to compare strings 

  Currently the tests on config generation run both the expected and actual config through a YAML serializer. This means that errors in the serialization (such as dropping a field) would not be caught. This is particularly important for fields that we want to include even when set to their zero value; these tests will not fail if that logic is implemented incorrectly.

  Also these tests re-implement the serialization logic, instead of calling the new GenerateConfigFiles.

  Let's fix the test to call GenerateConfigFiles, then compare the result directly against the expected result. It appears that the YAML is stable so the test shouldn't be flaky. (This is in contrast to protojson which deliberately has unstable output.)

  Because settings.js uses the output of protojson, and deck.env depends on the (non-deterministic) iteration order over a protobuf message fields, we'll omit those from the test.


maggieneterval (approved):
📈 🤝 🧪 💯

PR #95, opened 2020-06-15
feat(proto): mark AWS and Google canary accounts rootFolder as optional 

maggieneterval:
* feat(test): add tests for canary rootFolder fields 

  Add empty and non-empty rootFolder field to halconfig.yml. The empty rootFolder will be omitted from kayenta.yml, which is not the behavior we want. The next commit will mark rootFolder as optional since we are considering an empty string a valid value distinct from nil.

* feat(proto): mark AWS and Google canary accounts rootFolder as optional


ezimanyi (approved):
🥕 🧅 📁 💯

PR #96, opened 2020-06-16
refactor(proto): Wrap all boolean fields

ezimanyi:
We've decided that all boolean fields should have presence information.

Without this, booleans will not be included in the serialized config if they are 'false'. While this is fine for fields that default to false (for now), it breaks for any field that defaults to true. It also would cause issues if any field were changed in the future to default to true.

One of the advantages of kleat is that it only includes fields that are explicitly set by the user in its config, so it is easier to change field defaults, as only users who have explicitly set a field will be overriding the default.

But if we store a boolean without presence, it's impossible to ever in the future default it to 'true'. It also might be confusing for users that some fields in their hal config are not in their microservice configs.


internal/convert/orca.go
```diff
@@ -42,15 +42,15 @@ func getDefaults(h *config.Hal) *config.Orca_Defaults {
 
 func getPipelineTemplates(h *config.Hal) *config.Orca_PipelineTemplates {
 	// TODO: Consider whether this should actually be a BoolWrapper so that we
-	// can distinguish an explicitly set 'false' from an unset value
-	if !h.GetFeatures().GetPipelineTemplates() {
+	// can distinguish an explicitly set 'wrappers.False()' from an unset value
```
maggieneterval: Do we still need this TODO?

maggieneterval (approved):
🎁 👍 👎 🏑 💯

maggieneterval:
Maybe worth copying part of your PR description into the `CONTRIBUTING.md` to document that any new boolean fields added to kleat should be wrapped?

internal/convert/orca.go
ezimanyi: Oh, good call, I'll add a TODO to remove it.

ezimanyi: (Or will actually remove it...)

ezimanyi:
Added a commit to address the above comments:

* fix(core): Remove TODO and add documentation

Remove a TODO, and also fix the code which should pass through the
value when non-nil (not only when true).

Also document how to choose a field type. This also re-wraps the file.
I didn't otherwise change anythign above the added section, so didn't
break this into commits as it should be easy to see what was
re-wrapped vs what was added.

PR #97, opened 2020-06-16
feat(proto): Wrap string and int32 fields where necessary

ezimanyi:
* feat(kubernetes): Add missing Kubernetes fields

  A few fields were missing from KubernetesAccount. Add these.

* feat(proto): Wrap string and int32 fields where necessary

  This commit goes through all string and int32 fields, wrapping them where necessary. The only changes necessary were a few string fields.


ezimanyi:
Just enums left after this one!

api/proto/cloudprovider/kubernetes.proto
```diff
@@ -83,6 +83,51 @@ message KubernetesAccount {
...
+
+    // If set, all calls to kubectl will time out after the specified number of
+    // seconds.
+    int32 kubectlRequestTimeoutSeconds = 22;
+
+    // Whether to check whether the account has permission to read configured
+    // kinds before caching them. Kinds that the account does not have permission
+    // to read will be omitted from caching.
+    // This field defaults to true, and it is recommended to leave it at the
+    // default. If this field is set to false, any kubernetes objects that
```
maggieneterval:
```suggestion
    // default. If this field is set to false, any Kubernetes objects that
```

api/proto/cloudprovider/kubernetes.proto
```diff
@@ -83,6 +83,51 @@ message KubernetesAccount {
...
+    // service account to use.
+    string oAuthServiceAccount = 24;
+
+    // If true, only cache Kubernetes objects that have been deployed by
+    // Spinnaker, and ignore any other objects that exist in the cluster.
+    // Defaults to false.
+    google.protobuf.BoolValue onlySpinnakerManaged = 27;
+
+    // If true, enable detailed logging for all communications with the
+    // kubernetes cluster for this account. Defaults to false.
```
maggieneterval:
```suggestion
    // Kubernetes cluster for this account. Defaults to false.
```

maggieneterval: (just because we are capitalizing it everywhere else)

maggieneterval (approved):
🎁 🐍 🔢 ❓ ✅

PR #98, opened 2020-06-16
feat(proto): Add unspecified to enum fields

ezimanyi:
* style(test): Sort halconfig file

  It's easier to find fields if it's sorted.

  Also while making trivial changes, format the markdown files with prettier.

* feat(security): Remove unused enum

  This enum was only used by Halyard to track they type of OAuth provider you had configured; it used this to set other fields when you ran hal commands.

  A while back I updated the documentation on spinnaker.io to show users how to set the required fields themselves (which is useful both for directly editing your hal config with Halyard, or for early adopters of kleat).

  We can remove this field as it is not used. I'll add a note to the changelog about this. Existing Halyard users would already have had the other fields filled out based on that field when they ran hal commands, so I'm not including this as a breaking change. New users will need to set their config fields themselves using the documentation on spinnaker.io.

* feat(proto): Add unspecified to enum fields

  We want to distinguish unspecified enums from enums set to their default value. Let's do this by making the default value in each enum be a sentinel UNSPECIFIED value. This means that when the field is not specified, it will be deseriazlied as UNSPECIFIED, and will be omitted from re-serialization. Any valid value will be included in serialization.


maggieneterval (approved):
❓ 📧 #️⃣ 📝 ✅

PR #99, opened 2020-06-16
feat(ci): Remove default exclusions from golang-ci

ezimanyi:
* feat(ci): Remove default exclusions from golang-ci

  It turns out some things (including commenting exported functions) are turned off by default on golang-ci. (And I've already forgotten to comment three exported functions in the pas few days.) Let's turn off the default exclusions, as we should be starting from a clean slate.

  If it turns out there is something it flags that we find annoying and don't want to fix, we can figure out how to turn that off individually.

  I'm making this commit separately as I want to ensure that it catches the errors I think it should, then will fix in a follow-up commit.

* style(core): Add some missing comments, and fix package

  Add comments to the exported functions that are missing comments. Also, I realize that I used the older version of the proto library when writing the helper wrapper functions. This worked because wrappers in the old library just aliases BoolValue to the new library, but let's remove the indirection (and direct dependence on the new library).


internal/wrappers/wrappers.go
```diff
@@ -1,15 +1,20 @@
+// Package wrappers provides functions for working with wrapper messages
+// in the google.protobuf package.
 package wrappers
 
-import "github.com/golang/protobuf/ptypes/wrappers"
+import "google.golang.org/protobuf/types/known/wrapperspb"
 
-func True() *wrappers.BoolValue {
+// True returns a wrapper BoolValue containing the value true.
+func True() *wrapperspb.BoolValue {
```
maggieneterval, line 8: 🎁 🥜 🥪

maggieneterval (approved):
🎉

PR #100, opened 2020-06-17
feat(pkg): add monitoring daemon support

maggieneterval:
Adds monitoring dameon metric stores, hal to monitoring translation, and test coverage.


api/proto/config/monitoring.proto
```diff
@@ -0,0 +1,49 @@
...
+import "metricstores/datadog.proto";
+import "metricstores/newrelic.proto";
+import "metricstores/prometheus.proto";
+import "metricstores/stackdriver.proto";
+
+// Configuration for the spinnaker-monitoring microservice.
+// The monitoring protos use snake_case for backwards compatibility with
+// Halyard-generated hal configs and the spinnaker-monitoring microservice.
+// All new protos should be added using camelCase for consistency with the
+// rest of the hal config.
```
maggieneterval, line 15: Let me know if you think this makes sense. Alternatively, we could require users to supply monitoring config in camelCase and document it as a breaking change, but then translate these fields to snake_case since that is how spinnaker-monitoring expects them.

testdata/spinnaker-monitoring.yml
```diff
@@ -0,0 +1,27 @@
+datadog:
+  enabled: true
```
maggieneterval, line 2: Each individual metric store's `enabled` field is technically unnecessary to pass to spinnaker-monitoring since it is only using the `monitor.metric_store` list to determine which stores are enabled. That said, it seemed unnecessarily complex to explicitly remove `enabled` from each store in the output, either by manually nulling it out or by creating separate output protos for each store without that field.

ezimanyi (approved):
💯 💯 💯 💯 💯 💯 💯 💯 
(for the PR number)

📊 🏪 🥳 🥇 🎈 
(for the changes in this PR)

api/proto/config/monitoring.proto
ezimanyi: I think this makes sense, in the spirit of having `kleat` be as lightweight a translation layer as possible.

testdata/spinnaker-monitoring.yml
ezimanyi: Is one option here to just remove it from the proto entirely so that it gets ignored on input and output?

maggieneterval: Oh well we do need it in the input, then the translation layer writes each enabled type to `monitor.metric_store`.

ezimanyi: Oh, right of course, I just looked at that code.  I think we should leave this how you have it now!

PR #102, opened 2020-06-17
fix(core): Address ignored fields from test halconfig

ezimanyi:
* fix(test): Remove unused fields from test halconfig

  This commit removes fields that are not used by kleat from the test halconfig.

* feat(proto): Add some missing protos

  These are fields that Halyard defaults but doesn not expose commands for editing (though users can still manuall set them). In order to preserve backwards-compatibility for users migrating from Halyard we should still support reading them. Ideally the fields should be defaulted to something sensible in the service, but this will not affect migrating users who already have them specified.

* fix(proto): Fix mis-spelled requiredGroupMembership

  This was incorrectly included as requiredGroupMemberships in a number of protos.

* feat(proto): Remove experimental flag

  As we're not using optional protos, remove the experimental flag as we don't need it anymore.

* chore(core): Bump deps and run go mod tidy


testdata/halconfig.yml
```diff
@@ -190,13 +190,13 @@ metricStores:
     app_key: def
     enabled: true
     tags:
-      - key:val
+    - key:val
```
maggieneterval, line 193: Sorry! 🙈

maggieneterval (approved):
✉️ 🙈 🔢 ✅ ⚒️ 🐛 💯

PR #103, opened 2020-06-19
feat(stats): Include stats fields in echo config

ezimanyi:
* test(echo): Refactor echo tests to handle stats changes

  The stats block has a property that no other field in the hal config has; even when nothing is specified in the input, we still generate some output.

  This is because one of the blocks in the stats config is the deployment method, which kleat needs to fill in with details about itself. While this is a departure from the usual situation where empty input leads to empty output, it feels like this is the one case where an exception is warranted given that this output needs to be populated by kleat rather than by the user.

  To make the diff on that change smaller, refactor the tests now to include this default value for the Stats field (currently nil).

* refactor(stats): Use a different message for input/output stats

  While currently the only configurable stats field is whether it is enabled, we'll soon be adding more fields. Some fields will only be in the output (ie, that Spinnaker is being deployed by kleat and the kleat version) and should not be in the input. To allow us to do this, use a different message for the block in the hal config and in the echo config.

  Another approach would be to use the same message, but have the fields be ignored in the input. I didn't do that as I don't want those documented as they might add confusion, but am definitely open to further thoughts on this.

* feat(stats): Add stats protos

  Add the remaining fields to the stats protos. This commit does not use these new fields yet, but it does lightly refactor the tests to explicitly include some of the fields as nil in the tests (making the diff in the next commit cleaner).

* feat(stats): Include stats fields in echo config

  This commit includes the newly-added stats fields in the generated echo config, and adds some tests on these fields.


maggieneterval (approved):
📊 🔉 🔢 💯 

> Another approach would be to use the same message, but have the fields be ignored in the input. I didn't do that as I don't want those documented as they might add confusion, but am definitely open to further thoughts on this.

I agree with your approach! I think we've been reusing protos for input/output when the output doesn't necessarily need every field from the input (but should be transparent to the end-user), but in this case since it's the opposite I definitely agree with your approach to reduce confusion.

PR #104, opened 2020-06-22
feat(stats): Pass version to echo config

ezimanyi:
* refactor(core): Make version fields accessible outside of package

  This commit pulls the variables to hold build-time information into their own package and exports them (so that consumers don't need to import things from cmd).

* feat(stats): Pass version to echo config

  Pass the version of kleat used to generate the config to the echo config so that it is recorded along with the deployment method.


maggieneterval (approved):
⛹️‍♂️ 🏀 ➡️ ⛹️‍♀️ 🥇

PR #107, opened 2020-06-24
chore(core): Bump dependencies

ezimanyi:
* chore(core): Bump dependencies

  go get -u ./...

* chore(proto): Bump protoc-gen-go

  Also while updating the docker file, revert to using the official release of protoc-gen-doc as we don't need optional support anymore. This way we're no longer depending on the version in my fork.

* refactor(proto): Remove wrapper package

  google.golang.org/protobuf v1.25.0 adds new helper functions that obviate the need for the ones that we wrote. Let's remove our layer of indirection and just call the new functions.


maggieneterval (approved):
🔢 🎁 ✅ 💯 🏅

PR #108, opened 2020-06-25
fix(deck): create empty providers.kubernetes entry if provider is ena…

maggieneterval:
…bled

As explained in this [PR](https://github.com/spinnaker/deck/pull/8381), the Kubernetes provider does not have any statically configured defaults in Deck. However, the registration of the Kubernetes cloud provider in Deck depends on the truthiness of `providers.kubernetes`.


api/proto/config/deck.proto
```diff
@@ -161,6 +161,8 @@ message Deck {
     // Tencent Cloud provider defaults.
     TencentCloud tencentcloud = 9;
 
+    Kubernetes kubernetes = 10;
```
ezimanyi, line 164: If we always intend this to stay empty, this could instead be of type `google.protobuf.Empty` instead of defining a new type with no fields.

ezimanyi (approved):
💯 🥇 ✅

api/proto/config/deck.proto
maggieneterval: Whoah I didn't even know that existed! I guess I'd very slightly prefer to leave the 🚪 open to statically-configured Kubernetes-specific properties derived from the top-level Spinnaker config, since I think configuring this as `Empty` would be communicating that we made an informed decision to explicitly disallow this, which I don't think we want.

PR #109, opened 2020-06-29
Improve help message for generate command

achew22:
Update the `Use` parameter to include usage per the cobra docs:

```
    // Use is the one-line usage message.
```

This makes the output in the "Usage" section to contain a description of the mandatory arguments.

Additionally updates the usage example to contain the command name.

### DIFF
--- a/cmd/generate.go
+++ b/cmd/generate.go
@@ -23,7 +23,7 @@ import (
 )
 
 var generateCmd = &cobra.Command{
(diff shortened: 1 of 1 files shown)


ezimanyi (approved):
Looks good, thanks for this!

maggieneterval (approved):
Thanks so much! 🎉

PR #113, opened 2020-07-01
Create the output directory if it doesn't exist

### LINKED ISSUE #110: `kleat generate` expects that the output dir already exists
If `kleat generate` is invoked with a dir an output dir that doesn't exist, you get the following error.

```
Error: stat ./generated: no such file or directory
```

Proposed solutions:

1. Notify the user that they need to create that directory
2. Make that directory for the user
3. Something else I didn't think of?

Which would you prefer in the PR?

achew22:
Fixed: #110


internal/fileio/fileio.go
```diff
@@ -36,16 +36,16 @@ import (
 func ParseHalConfig(halPath string) (*config.Hal, error) {
 	data, err := ioutil.ReadFile(halPath)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("unable to read %q error: %v", halPath, err)
```
maggieneterval:
```suggestion
		return nil, fmt.Errorf("unable to read %q: %v", halPath, err)
```
or:
```suggestion
		return nil, fmt.Errorf("unable to read %q. Error: %v", halPath, err)
```
No strong opinion other than these be relatively consistent and the actual error message is cleanly separated from the custom message.

maggieneterval (approved):

internal/fileio/fileio.go
maggieneterval: @achew22 I ended up just committing one of these suggestions so there's some separation between the error message, so that we can get this merged. Thanks again for the contribution!

achew22:
Sorry, I completely forgot about the PR in my email deluge. Thanks for cleaning up after me!

PR #114, opened 2020-07-07
feat(authn/saml): New optional SAML parameter for signing messages

german-muzquiz:
Related Gate PR: https://github.com/spinnaker/gate/pull/1269
Related Halyard PR: https://github.com/spinnaker/halyard/pull/1741


api/proto/security/authn/authn.proto
```diff
@@ -187,6 +190,27 @@ message Saml {
     // Email.
     string email = 6;
   }
+
+  // Digest algorithms to sign SAML messages.
+  enum SignatureDigest {
+    // Digest algorithm SHA1 (default).
```
ezimanyi, line 199: In general we're recommending that all `enum`s `UNSPECIFIED = 0` as their first (default) entry.  There are more details [here](https://github.com/spinnaker/kleat/blob/1754772f46c6c8957cb68e8832e84a2862548eb4/CONTRIBUTING.md#choosing-field-type), but the main reason is that enums set to their default will be omitted from serialization.  (So if a user explicitly sets the signature digest to SHA1, it won't be included in the output YAML if that's the default.)

In this case, since Gate uses `SHA1` as the default it wouldn't matter, because there would be no difference to gate between `signatureDigest: SHA1` and omitting `signatureDigest`.  But then this creates a coupling between `kleat` and `gate` where the correctness here depends on the defaulting in `gate` (and if we ever change the default in `gate` this will break).

ezimanyi (commented):
Looks great!  Just one comment about adding an explicit `UNSPECIFIED` to the enum.  Thanks for this!

ezimanyi (approved):
Looks great, thanks for this!

PR #115, opened 2020-07-08
fix(protoc): Run protoc container as invoking user

ezimanyi:
* style(proto): Add newlines in docker command 

  This will make it easier to see what changed when adding/removing arguments to the command.

* refactor(proto): Use mount instead of volume command 

  The mount command is newer and recommended by Docker. It also allows us to mount a volume readonly, which we want to do for the proto directory (which is input only).

* refactor(protoc): Use better directories in container 

  It's customary for mounted filesystems to be in /mnt; let's mount our files there instead of directly in the root of the filesystem.

  Rather than have a top-level staging directory, make this directory in /tmp. Also create it as part of the genproto script so it's automatically owned by whichever user is running the script.

* fix(protoc): Fix file permissions in kleat-protobuilder 

  This commit updates the intermediate containers used to fetch protoc and its dependencies to no longer run as root.

  The first change is to update the 'curl' container to run as a non-root user 'curl' to set the work directory to this user's home directory.

  After downloading and unzipping the desired files, we run chmod to set the desired permissions, which in many cases are not set to what we want in the tar/zip archive. We want 755 for directories and executable files and 644 for non-executable files, so that only the owner can modify but anyone can use the files when running protoc.

  When copying the files to the kleat-protobuilder container, explicitly have them owned by root (leaving the permissions as what we set them to after downloading). This matches the way things are normally installed on linux, where root owns the files and is the only user with permission to write, but any user can read/execute the files.

  In the case of genproto, we'll explicitly follow up with a chmod to ensure it has mode 755 as we can't necessarily control what the user has on their system.

  As of this commit, we've fixed the permissions so that it is now possible to run the kleat-protobuilder as a non-root user by passing the -u flag to docker, but we're not actually doing that yet, which will follow in an upcoming commit.

* fix(protoc): Run protoc container as invoking user 

  This commit adds the -u flag to the docker command running the protocol buffer compilation, so that the container runs with the same uid/gid as the invoking user. This will cause any output files to have this uid/gid (instead of being owned by root as before).

  I've also created a user 'protoc' in the container and set the user to that user. This is really just a fallback so we're falling back to a non-root user; this will always be overridden by the --user flag that we've added to the docker run command.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,9 +3,9 @@ protobuilder:
 
 proto_command = docker run \
   --rm \
-  --mount type=bind,source=$(abspath api/proto/),target=/proto,readonly \
-  --mount type=bind,source=$(abspath api/client/),target=/output/go \
-  --mount type=bind,source=$(abspath docs),target=/output/doc \
+  --mount type=bind,source=$(abspath api/proto/),target=/mnt/proto,readonly \
+  --mount type=bind,source=$(abspath api/client/),target=/mnt/output/go \
+  --mount type=bind,source=$(abspath docs),target=/mnt/output/doc \
   kleat-protobuilder:latest
 
 proto: protobuilder
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,12 @@
 protobuilder:
 	docker build -q -t kleat-protobuilder:latest build/protoc/
 
-proto_command = docker run --rm -v $(abspath api/proto/):/proto -v $(abspath api/client):/output/go -v $(abspath docs):/output/doc kleat-protobuilder:latest
+proto_command = docker run \
+  --rm \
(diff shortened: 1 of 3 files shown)


ezimanyi:
It looks like MacOs does some translation, so that even when the container is running as root, the files it writes to mounted volumes are still owned by the user who invoked `docker`, which is why I didn't notice this before.  I ran `make proto` from my Linux machine recently and noticed that all the files were owned by `root` which was annoying as then I couldn't overwrite them without `sudo`.

maggieneterval (approved):
👨‍💻 ⚒️ 💻 🔐 🌳 ⬇️ 🚫 ✅ 💯 🤯

PR #118, opened 2020-07-17
docs(api): Edit Kleat API documentation for Bitbucket artifact config.

dorbin:


maggieneterval:
@dorbin I think you'll need to [run `make proto`](https://github.com/spinnaker/kleat/blob/master/CONTRIBUTING.md#regenerating-dependent-code) to generate the protos, which is why CI is failing.

ezimanyi (approved):
Looks good, thanks!

PR #119, opened 2020-07-19
fix(core): remove dups 'reading' on fileio package top level synopsis

zchee:
Remove dups 'reading' on fileio package top-level synopsis.


ezimanyi (approved):
Thank you!

PR #120, opened 2020-07-19
fix(all): run goimports happy

zchee:
fix(all): run goimports happy.

```console
$ goimports -w -local=github.com/spinnaker/kleat .
```

PR #122, opened 2020-07-20
feat(core): Add local-prefixes flag to golint-ci

ezimanyi:
Now that we've cleaned up our imports to be nicely grouped, with local imports last, add the same flag to golint-ci so that future PRs respect this style.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -10,3 +10,5 @@ linters:
     - goimports
     - golint
     - govet
+goimports:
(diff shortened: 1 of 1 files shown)


ezimanyi:
I expect the CI to fail until this is rebased onto #120.

maggieneterval (approved):

ezimanyi:
(Still trying to figure out why the CI is passing here...)

PR #123, opened 2020-07-20
feat(dependencies): Add dependabot config

ezimanyi:
Let's use dependabot to keep our go dependencies up to date, as well as the base images in our Dockerfile for compiling protos.

This will *not* upgrade the versions of the following in our build
container, which we'll still need to do manually:
* protoc
* protoc-gen-go
* protoc-gen-doc

But it still seems worthwhile to have some of these dependencies auto-bumped for us.

### DIFF
--- a/dependabot.yml
+++ b/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+- package-ecosystem: "gomod"
+  directory: "/"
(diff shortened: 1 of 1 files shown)


maggieneterval (approved):
👭 🤖 ✅

PR #127, opened 2020-07-28
Edit kleat canary protos

dorbin:


dorbin:
@fieldju

api/proto/canary/canary.proto
```diff
@@ -17,25 +17,25 @@ message Canary {
...
+  // Whether to show _all_ canary configs in Deck, or just those scoped to
   // the current application.
   google.protobuf.BoolValue showAllConfigsEnabled = 5;
 
-  // Whether or not to enable custom filter templates for canary configs in Deck.
+  // Whether to enable custom filter templates for canary configs in Deck.
   google.protobuf.BoolValue templatesEnabled = 6;
 
-  // The default canary judge. Defaults to `NetflixACAJudge-v1.0`, which is
+  // The default canary judge. Defaults`NetflixACAJudge-v1.0`, is
```
ezimanyi: This seems less clear after the edit.

ezimanyi (approved):

api/proto/canary/canary.proto
dorbin: Yeah, not sure what I did there.

api/proto/canary/google.proto
```diff
@@ -32,32 +32,36 @@ message GoogleAccount {
...
 
-  // The root folder in the chosen bucket in which to store all of the canary
-  // service's persistent data in. Defaults to `kayenta`.
+  // The root-level folder, in the specified bucket, in which to store all
+  // the canary service's persistent data. Defaults to `kayenta`.
   google.protobuf.StringValue rootFolder = 5;
 
   // (Required) The Google Cloud Platform project the canary service will use to
-  // consume GCS and Stackdriver.
+  // consume telemetry.
```
maggieneterval: Does this make it less clear that the project could also be used for GCS in addition to Stackdriver-supplied telemetry data?

api/proto/canary/aws.proto
```diff
@@ -40,14 +40,17 @@ message AwsAccount {
...
+  // The endpoint used to reach the service implementing the S3 API. Typically
+  // you would use this with an S3 clone, like Minio.
   string endpoint = 6;
 
   // The default access key used to communicate with AWS.
   string accessKeyId = 7;
 
-  // If enabling S3, include CONFIGURATION_STORE and/or OBJECT_STORE in this
-  // list.
+  // Use METRICS_STORE for AWS monitoring.
```
maggieneterval: This might have been correct as-is; not actually sure we support an AWS metrics store based on our [docs](https://spinnaker.io/setup/canary/#identify-your-metrics-provider)?

dorbin: But they'd want to include `METRICS_STORE`, regardless what metrics provider they use, right?

maggieneterval: I don't think so, I think you'd only want to specify `METRICS_STORE` for accounts for integrations that could be used to supply metrics. I think the AWS integration is only for S3 as a persistent store, for either (or both) of `CONFIGURATION_STORE` and `OBJECT_STORE`.

fieldju:
~If this is for canary config, I have documented the API spec for this with MSON here:~
~https://github.com/spinnaker/kayenta/blob/master/docs/canary-config.md~

[edit]
I would cross-reference this: 
https://github.com/spinnaker/kayenta/blob/master/kayenta-web/config/kayenta.yml

We have attempted to list all configuration options there.

api/proto/canary/aws.proto
fieldju: https://github.com/spinnaker/kayenta/blob/master/kayenta-web/config/kayenta.yml#L57

fieldju: Only has object store and configuration store listed.

api/client/canary/aws.pb.go
```diff
@@ -32,7 +32,7 @@ type Aws struct {
 	sizeCache     protoimpl.SizeCache
 	unknownFields protoimpl.UnknownFields
 
-	// Whether the integration is enabled.
+	// Whether the AWS canary integration is enabled.
```
fieldju: Calling it `AWS Canary` integration is confusing to me.

Really using this allows Kayenta to store canary config and archived results in an S3 bucket.

api/client/canary/aws.pb.go
```diff
@@ -93,7 +93,7 @@ func (x *Aws) GetS3Enabled() *wrappers.BoolValue {
 	return nil
 }
 
-// Configuration for an AWS account.
+// Configuration for an AWS canary account.
```
fieldju: Same comment as above.

The use of `canary` here doesn't make sense.

api/client/canary/aws.pb.go
```diff
@@ -113,13 +113,16 @@ type AwsAccount struct {
...
-	// The endpoint used to reach the service implementing the AWS API. Typical
-	// use is with Minio.
+	// The endpoint used to reach the service implementing the S3 API. Typically
+	// you would use this with an S3 clone, like Minio.
 	Endpoint string `protobuf:"bytes,6,opt,name=endpoint,proto3" json:"endpoint,omitempty"`
 	// The default access key used to communicate with AWS.
 	AccessKeyId string `protobuf:"bytes,7,opt,name=accessKeyId,proto3" json:"accessKeyId,omitempty"`
-	// If enabling S3, include CONFIGURATION_STORE and/or OBJECT_STORE in this
-	// list.
+	// Use METRICS_STORE for AWS monitoring.
```
fieldju: Same comment as @maggieneterval 
AWS doesn't have `METRICS_STORE` as an option.

This would likely come as its own CloudWatch integration.
See: https://github.com/spinnaker/kayenta/issues/757

api/client/canary/canary.pb.go
```diff
@@ -34,20 +34,20 @@ type Canary struct {
 
 	// Whether the canary service is enabled.
 	Enabled *wrappers.BoolValue `protobuf:"bytes,1,opt,name=enabled,proto3" json:"enabled,omitempty"`
-	// Canary service integrations. You must configure at least one account of
-	// each canary.SupportedType (METRICS_STORE, CONFIGURATION_STORE,
-	// OBJECT_STORE) in order to use Spinnaker's canary service.
+	// Canary service integrations. To use Spinnaker's automated canary service,
```
fieldju: You could even link to this for more information:
https://github.com/spinnaker/kayenta/blob/master/docs/configuring-kayenta.md

Not sure if adding `To use Spinnaker's automated canary service,` makes things clearer to me or more confusing.

There isn't really a canary service.

`To enable Spinnaker's Automated Canary Analysis (ACA) features` makes more sense to me.

ACA is a user experience that spans 4 services (Deck, Deck-Kayenta, Orca, and Kayenta)

api/client/canary/datadog.pb.go
```diff
@@ -32,9 +32,9 @@ type Datadog struct {
 	sizeCache     protoimpl.SizeCache
 	unknownFields protoimpl.UnknownFields
 
-	// Whether the integration is enabled.
+	// Whether the Datadog canary service integration is enabled.
```
fieldju: `Canary service` is a bad choice of words here IMO.

We should try to be consistent with our language choices and speak a ubiquitous language around the domain of Automated Canary Analysis (ACA).

We call these `Metrics Stores` in existing documentation.

api/client/canary/newrelic.pb.go
```diff
@@ -32,7 +32,7 @@ type NewRelic struct {
 	sizeCache     protoimpl.SizeCache
 	unknownFields protoimpl.UnknownFields
 
-	// Whether the integration is enabled.
+	// Whether the New Relic canary integration is enabled.
```
fieldju: `canary integration` -> `metric store`

api/client/canary/prometheus.pb.go
```diff
@@ -32,11 +32,11 @@ type Prometheus struct {
 	sizeCache     protoimpl.SizeCache
 	unknownFields protoimpl.UnknownFields
 
-	// Whether the integration is enabled.
+	// Whether the Prometheus canary integration is enabled.
```
fieldju: canary integration -> metric store

api/client/canary/s3.pb.go
```diff
@@ -27,15 +27,15 @@ const (
...
+// also configure at least one `canary.AwsAccount` with a list of
+// `supportedTypes` that includes `canary.SupportedType.CONFIGURATION_STORE`
+// and/or `canary.SupportedType.OBJECT_STORE`.
 type S3 struct {
 	state         protoimpl.MessageState
 	sizeCache     protoimpl.SizeCache
 	unknownFields protoimpl.UnknownFields
 
-	// Whether the integration is enabled.
+	// Whether the S3 canary integration is enabled.
```
fieldju: Not sure how to re-word this, because this has to be used with `kayenta.aws.accounts` 
And its to enable a storage engine more than it is a `canary intagration`

api/client/canary/signalfx.pb.go
```diff
@@ -32,7 +32,7 @@ type SignalFx struct {
 	sizeCache     protoimpl.SizeCache
 	unknownFields protoimpl.UnknownFields
 
-	// Whether the integration is enabled.
+	// Whether the SignalFx canary integration is enabled.
```
fieldju: `canary integration` -> `metric store`

fieldju:
I stopped leaving the same comment over and over, but I find it confusing that Kaytena metric, object, and configuration stores are all just generalized and referred to as `canary integrations`

Maybe it's fine, but I'd like it if we took a more Domain-Driven Design (DDD) approach and had a shared set of words that had the same meaning across projects and teams (ubiquitous language).

api/proto/canary/aws.proto
dorbin: :+1: 
Thanks, Maggie and Justin!

dorbin:
@fieldju , I've updated per your suggestions, hopefully haven't missed any. PTAL when you get a chance. Thanks!

dorbin:
@plumpy

api/proto/canary/prometheus.proto
```diff
@@ -7,36 +7,35 @@ import "google/protobuf/wrappers.proto";
...
 
-  // Number of milliseconds to wait in between caching the names of available
+  // Number of milliseconds to wait between caching the names of available
   // metric types (used when building canary configs). Defaults to 60000.
   int32 metadataCachingIntervalMS = 3;
 }
 
 // Configuration for a Prometheus account.
+// For authentication, you must provide either `usernamePasswordFile` or
+// `username` and `password`
```
plumpy:
```suggestion
// For authentication, you must provide either `usernamePasswordFile` or
// `username` and `password`.
```

api/proto/canary/stackdriver.proto
```diff
@@ -5,15 +5,15 @@ option go_package = "github.com/spinnaker/kleat/api/client/canary";
 
 import "google/protobuf/wrappers.proto";
 
-// Configuration for the Stackdriver canary integration. If enabled, you must
-// also configure at least one canary.GoogleAccount with a list of
-// supportedTypes that includes canary.SupportedType.METRICS_STORE.
+// Configuration for the Google Cloud Monitoring canary integration. If
```
plumpy: do you want to add "(formerly Stackdriver)" here?

plumpy (approved):

PR #128, opened 2020-08-03
Allow for multiple halconfigs as input

dsexton:
This will allow for multiple YAML files to be merged prior to being fully parsed, and should not break any current usage of the command.


ezimanyi:
Thanks for opening this PR!  Meant to take a look today but ran out of time---will take a look in the next few days.  Thanks!

dsexton:
@ezimanyi no problem. I kept the original CLI arg parsing logic so any existing scripts leveraging kleat will not need to be modified. 

However, I would suggest adopting flags in the future. Maybe it’s subjective, but I find it better to capture arguments of differing domains as flags, or mostly flags with a single domain as arguments. This also makes the auto generated command documentation a bit cleaner. Just my two cents anyway.

internal/fileio/fileio.go
```diff
@@ -39,13 +41,37 @@ func ParseHalConfig(halPath string) (*config.Hal, error) {
...
+// ParseHalConfigs reads the YAML files at halPaths parses it into a *config.Hal.
+// The config.Hal is validated; if there are any errors, they will be returned
+// in error and *config.Hal will be nil.
+func ParseHalConfigs(paths []string) (*config.Hal, error) {
+	merged, err := mergeConfigs(paths)
+	if err != nil {
+		return nil, err
+	}
+
+	data, err := yaml.Marshal(merged)
```
ezimanyi, line 56: I wonder if there's a way to do this without having to round-trip to/from YAML.

internal/fileio/fileio.go
```diff
@@ -85,3 +111,43 @@ func ensureDirectory(d string) error {
...
+			return nil, fmt.Errorf("unable to parse %q: %w", halPath, err)
+		}
+
+		base = mergeMaps(base, currentMap)
+	}
+
+	return base, nil
+}
+
+func mergeMaps(a, b map[string]interface{}) map[string]interface{} {
```
ezimanyi, line 136: I guess this is a fairly simple function, so maybe the benefit of pulling it in from a library might be questionable...but is this something that there might be a library for?  In the absence of that some tests on it would probably be good.

ezimanyi (commented):
Thanks a lot for this PR!  I left a couple of tiny comments, but I think overall the approach is good.

I actually like your idea of adding CLI flags; one reason is that part of me wonders if in the future we find a better solution for secret management, we may want to encourage that and move people away from this solution.  Having this as a separate flag would probably make that easier (and in general it is less risky to make changes to a sub-command than to our root command).

internal/fileio/fileio.go
dsexton: I actually borrowed this from Helm's merge implementation, let me know if this requires any additional citing.

internal/fileio/fileio.go
dsexton: Should we leave a TODO to explore alternatives?

ezimanyi: It looks like there's a [Merge](https://pkg.go.dev/google.golang.org/protobuf/proto?tab=doc#Merge) function generated for types `proto.Message` that we would be able to use here.   (Per the godoc for that function, you can also just repeatedly unmarshal into the same proto specifying the `Merge` strategy.)

I think that would be preferable as it uses a well-documented upstream merge strategy that uses the type information we have on our messages.

ezimanyi:
I left a comment about using the `Merge` function that we get for free from using the `proto` library.  Also I guess the two additional things that I'd like to see before merging this:
* Some test coverage ensuring that we are properly merging two hal configs (as none of the existing test cases cover that and I'm afraid it will break otherwise)
* Moving the multiple-config to its own parameter, just to provide some isolation while we test it out and get feedback.  (While this is still new, I'm a bit hesitant to change the default base command without some feedback.)

ezimanyi:
Closing due to age; feel free to re-open if you pick this up again.

PR #129, opened 2020-08-05
refactor(proto): exclude "mine canary" feature flag from Deck's protos

maggieneterval:
Related to: https://github.com/spinnaker/spinnaker/issues/5939

Upon further investigation, the "mine canary" feature flag should only be gating the pre-Kayenta legacy canary stages, which we suspect is currently only used by opsmx's custom build. To avoid confusing OSS users, do not permit configuration of this flag with kleat.


api/proto/features.proto
```diff
@@ -7,12 +7,12 @@ import "google/protobuf/wrappers.proto";
 
 // Feature flags
 message Features {
+  reserved 2;
+  reserved "mineCanary";
```
maggieneterval: Definitely want to talk about this since I think it's the first field we've removed post-alpha-release. Looks like the choices are:

1. [Add the `deprecated` proto annotation.](https://developers.google.com/protocol-buffers/docs/proto3#options)
2. [Remove the field and add `reserved` name and number.](https://developers.google.com/protocol-buffers/docs/proto3#reserved)

Went with (2) since I was completely wrong about this field, and it's never been necessary (as far as I can tell) to configure non-legacy OSS canarying. (Until, of course, I introduced a bug in 1.21 where this legacy flag gates non-legacy canary data sources, which I will fix shortly...).

Before our first GA release we'll come up with a plan for communicating fields that have changed or are no longer necessary and require user intervention.

maggieneterval: I guess a third option would be to just delete the field and re-number, which I _think_ wouldn't be super problematic (yet) as we're only using these protos to generate a translation layer?

As a second example, the other field I was planning on removing was Deck's `changelog` field since as of 1.21 (which I think is the earliest release we're recommending Kleat be used with) [it's no longer read by Deck](https://github.com/spinnaker/deck/pull/8353).

ezimanyi: Thanks for the great explanation of the options here!  I think either (2) or (3) would be best.  [The one advantage I can think of for (1) is it could allow us to leave some documentation about why the field was deprecated for a while; but I don't think that's relevant in this particular case.]

I agree that since we (at least at this point) are not guaranteeing that serialized protos will be backwards-compatible and are just using them as a translation layer that deleting the field won't actually cause issues.  But maybe we should just make (2) the plan going forward anyway as I don't think that there's much harm in reserving names/numbers.

Maybe we can do a final cleanup before GA where we actually commit to backwards compatibility?

internal/convert/deck.go
```diff
@@ -78,7 +78,6 @@ func getDeckChangelogConfig(_ *config.Hal) *config.Deck_Changelog {
 func getDeckFeaturesConfig(h *config.Hal) *config.Deck_Features {
 	return &config.Deck_Features{
 		PipelineTemplates:            h.GetFeatures().GetPipelineTemplates(),
-		Canary:                       h.GetFeatures().GetMineCanary(),
```
ezimanyi, line 71: Does this mean we can also remove the `canary` field from the `Deck_Features` proto?
https://github.com/spinnaker/kleat/blob/1d42fc37545346fb0f26b63c532c203540531e49/api/proto/config/deck.proto#L57

ezimanyi (approved):
🚫 ⛏️ 🐦 🎏 💯 ✅ 🥳

api/proto/features.proto
maggieneterval: I'm actually learning toward (3) at this point, and making it explicit that kleat doesn't guarantee backwards-compatibility of serialized protos. Otherwise we might end up with a graveyard of deprecated fields whose existence implies that we're guaranteeing or somehow relying on that backwards-compatibility. There's also no advantage to the end-user as the treatment of unknown fields will be the same regardless of whether the field number/name was retired, so folks are still going to need to read release notes to find out what fields are safe to remove.

internal/convert/deck.go
maggieneterval: Great catch, thank you! 🤗

api/proto/features.proto
ezimanyi: Yeah, I agree---let's go with (3)!

PR #130, opened 2020-08-05
feat(proto): remove unread Deck changelog proto

maggieneterval:
As of 1.21, Deck no longer requires config to render a changelog modal ([source](https://github.com/spinnaker/deck/pull/8353)).


api/proto/config/deck.proto
```diff
@@ -14,6 +14,9 @@ import "notification/slack.proto";
 import "notification/twilio.proto";
 
 message Deck {
+  reserved 6;
```
maggieneterval: (Happy to change this to however we decide to handle pre-GA field removals, per https://github.com/spinnaker/kleat/pull/129#discussion_r465915628), but had this on a branch and didn't want to forget about it.)

ezimanyi (approved):
🚫 💰 🪓 🌳 🥇

api/proto/config/deck.proto
maggieneterval: Removed reserved statement & updated field numbering.

PR #131, opened 2020-08-18
docs(kleat): Link to instructions in kustomization-base repo

ezimanyi:
Users landing on the kleat repo often have trouble finding the detailed install instructions we put in the kustomization-base repo. Let's link to them from the kleat README.

Maybe at some point these instructions should live in the kleat repo as most users looking up the tool will land here, but at least for now the link should help.


maggieneterval (approved):
🔗 📜 💻 ☁️ ✅ 
Thanks for thinking of this!

PR #132, opened 2020-08-20
Edit kleat ci docs

dorbin:


api/proto/ci/concourse.proto
```diff
@@ -30,7 +30,7 @@ message ConcourseAccount {
   // (Required) The URL at which your Concourse search is reachable.
   string url = 4;
 
-  // Fiat permissions configuration. A user must have at least one of the READ
+  // Configuration for Fiat permissions. A user must have at least one of the READ
```
maggieneterval, line 33: Trying to derive a rule from the changes here...when should we say "configuration for" vs. "configuration to use"?

maggieneterval (approved):

maggieneterval:
Also I think you might need to re-run `make proto` to get CI to pass.

api/proto/ci/concourse.proto
dorbin: No rule here; this was just my "ear." :ear:

Issue #136, opened 2020-09-03
Kleat missing default Kubernetes account in Deck's "settings.json"


KeisukeYamashita:
## What

When `kleat` generates Deck's `setting.js`, it is missing Kubernetes default configuration.

## Which version

* `kleat`: `0.4.0`
* Spinnaker: `1.20.8`

## How to reproduce

```consol
$ kleat generate <any_halconfig>
(see `settings.json`)
```

`kleat` will generate a file like this.

```
window.spinnakerSettings = JSON.parse('{"gateUrl":"https://spinnaker-api.something.domain.com","authEnabled":true,"authEndpoint":"https://spinnaker-api.something.domain.com/auth/user","bakeryDetailUrl":"https://spinnaker-api.something.domain.com/bakery/logs/{{context.region}}/{{context.status.resourceId}}","canary":{"defaultJudge":"NetflixACAJudge-v1.0","featureDisabled":false,"showAllConfigs":true,"templatesEnabled":true},"notifications":{"slack":{"enabled":false,"botName":"spinnakerbot","token":"${SLACK_TOKEN}"}},"providers":{"kubernetes":{}},"defaultTimeZone":"America/Los_Angeles","feature":{"pipelineTemplates":true,"chaosMonkey":false,"fiatEnabled":true,"managedPipelineTemplatesV2UI":true}}');
```

Here, the `provider.kubernetes` is `{}`. But it should be `provider.kubernetes: {default:{}}` to define the `default` property which the deck requires.  If I apply the manifests generated by kleat, it will run into error in the deck.

![image](https://user-images.githubusercontent.com/23056537/92116070-4c0c4400-ee2e-11ea-9c3e-768f4967bb20.png)

## Workaround

I'm currently configuring this manually is `settings-local.json`.

```console
$ echo "window.spinnakerSettings.providers.kubernetes = { defaults: {} }" >> settings.json
```

maggieneterval:
@KeisukeYamashita thanks for taking the time to open this issue! What version of Spinnaker are you trying to deploy? We're currently recommending 1.21+ since before that, many defaults were supplied by Halyard rather than the microservices themselves, so Kleat will not work as expected (we've documented that [here](https://github.com/spinnaker/kustomization-base#set-the-spinnaker-version), but should definitely highlight this more.). I think in this case, only the legacy Kubernetes (v1) provider required these provider defaults, and was only supported through 1.20. Let me know if I'm missing something, though!

[...]

PR #137, opened 2020-09-03
fix(deck): missing default for Kubernetes provider

KeisukeYamashita:
## What

Implemented a `default` field to Kubernetes Provider for Spinnaker Deck.

Is `default` field(=property is deck) is read by Deck. If we don't implement this, we have to configure in `settings-local.js` as below:

```
window.spinnakerSettings.providers.kubernetes = { defaults: {} }
```

It would be great if this will be fixed. With Halyard, we don't need to configure this.

## Why

* Solves #136


KeisukeYamashita:
This only happens in `1.20` release which `kleat` doesn't support. 
So, I will close this. Sorry for your confusion.

PR #138, opened 2020-09-16
docs(core): Add philosophy section to readme

ezimanyi:
Let's add a section outlining some of the design goals so that users and contributors understand the philosophy behind kleat and where it has deliberately departed from some design decisions that Halyard made.


maggieneterval (approved):
🤔 💭 ➡️ 📖 📚 💯 🎉 🥇

PR #143, opened 2020-09-29
feat(storage): add sql to storage options

nimakaviani:
Enabling the new datasource providers for redis and sql to be used for Keel


api/proto/storage/sql.proto
```diff
@@ -0,0 +1,40 @@
...
+      string user = 1;
+      // Database password
+      string password = 2;
+      // Database connection string. This needs to include server port
+      // and database name as well
+      string jdbcUrl = 3;
+      // Database connection timeout
+      int32 connectionTimeout = 4;
+      // maxLifetime controls the maximum lifetime of a connection in the pool.
+      int32 maxLifetime = 5;
```
maggieneterval, line 33: Is the unit for `connectionTimeout` and `maxLifetime` seconds? Might be worth clarifying.

api/proto/storage/sql.proto
```diff
@@ -0,0 +1,40 @@
+syntax = "proto3";
+package proto.storage;
+
+option go_package = "github.com/spinnaker/kleat/api/client/storage";
+
+import "google/protobuf/wrappers.proto";
+
+message SQL {
```
maggieneterval, line 8: I see that the `Redis` and `SQL` protos aren't actually consumed by any microservice config protos yet -- could you remind me if they are only consumed by Keel and the plan is to add Keel support in a followup PR?

maggieneterval (commented):
Thanks, @nimakaviani! Left a couple of questions.

api/proto/storage/sql.proto
nimakaviani: they are in milliseconds. updated the comments.

api/proto/storage/sql.proto
nimakaviani: Keel requires SQL. you are right that Redis isnt immediately needed. got rid of it.

api/proto/storage/sql.proto
maggieneterval: 👍  Thanks!

api/proto/storage/sql.proto
nimakaviani: Good point. I know that SQL can be used as an alternative to Redis for other microservices but I am not sure how well tested or supported it is. Maybe creating an issue to track the inconsistency is a better short term strategy. In our settings we use SQL only for Keel and have Redis as the datastore for other microservices. If you want, I can drop an issue for SQL vs Redis in other microservices.

maggieneterval: I think it's well-supported and might even be the recommended backend for Clouddriver and Orca at this point, but at least for Clouddriver there are other [parameters](https://spinnaker.io/setup/productionize/persistence/clouddriver-sql/#configure-clouddriver-to-use-mysql) that we'd need to add before fully supporting it in Kleat, and I don't want to block your addition of Keel support on that. If you want to add a TODO here or make a GH issue just to make it clear to future contributors that more work is needed to configure the SQL backend for the other microservices, that would be 💯 !

nimakaviani: done. Added issue https://github.com/spinnaker/kleat/issues/144

maggieneterval: Thanks so much, @nimakaviani, I really appreciate it! I'll keep an eye out for your next PR adding Keel support!

maggieneterval (approved):

PR #145, opened 2020-10-05
feat(keel): add support for ManagedDelivery

nimakaviani:
- add keel configuration
- updated other services when keel configuration is provided
- update tests with keel configuration


internal/convert/echo.go
```diff
@@ -33,6 +33,19 @@ func HalToEcho(h *config.Hal) *config.Echo {
...
+}
+
+func getEchoServices(h *config.Hal) *config.Echo_Services {
+	if !h.GetManagedDelivery().GetEnabled().GetValue() {
+		return nil
+	}
+
+	return &config.Echo_Services{
+		Keel: &config.ServiceSettings{
+			Enabled: h.GetManagedDelivery().GetEnabled(),
```
ezimanyi: Rather than including this block in the several places that need it, would it be possible to have users set `services.keel.enabled` in their service discovery file as documented for other services [here](https://github.com/spinnaker/kustomization-base#enable-optional-services).

When we looked at the first two optional services (`fiat` and `kayenta`) we leaned towards not having `kleat` do too much overriding of the service discovery file (the `spinnaker.yml`) figuring that that was more of a deployment concern.  (If you are enabling disabling services you'll also need to actually deploy/undeploy those.)

I think there are probably some improvements we can make to streamline the workflow, but as a first pass I'd lean towards not having `kleat` write out the `services.*` block to the service config.  Does that sound reasonable?

api/proto/config/deck.proto
```diff
@@ -44,6 +44,9 @@ message Deck {
   // Configuration for UI-related feature flags.
   Features feature = 10;
 
+  // Configuration for ManagedDelivery
```
ezimanyi: I think this should be "managed delivery" in the comments here (and in a few places below).

I'm only being particular about the comments because they generate user-facing documentation.

api/proto/config/managed_delivery.proto
```diff
@@ -0,0 +1,113 @@
...
+
+message Eureka {
+    // Whether this persistent store is enabled.
+    google.protobuf.BoolValue enabled = 1;
+}
+
+message KeelConfig {
+    // Managed Delivery plugins to be enabled
+    Plugins plugins = 1;
+    // duration string format (eg. 10s)
```
ezimanyi: This and the next comment should also briefly describe the field (as they are used in documentation).

api/proto/config/managed_delivery.proto
```diff
@@ -0,0 +1,113 @@
...
+    ResourceCheck resourceCheck = 2;
+    // in ISO 8601 format
+    ArtifactRefresh artifactRefresh = 3 [json_name = "artifact-refresh"];
+    // constraints pertinent to actions to be done to the pipeline
```
ezimanyi: s/constraints/Constraints/

api/proto/config/managed_delivery.proto
```diff
@@ -0,0 +1,113 @@
...
+    google.protobuf.BoolValue enabled = 1;
+}
+
+message KeelConfig {
+    // Managed Delivery plugins to be enabled
+    Plugins plugins = 1;
+    // duration string format (eg. 10s)
+    ResourceCheck resourceCheck = 2;
+    // in ISO 8601 format
+    ArtifactRefresh artifactRefresh = 3 [json_name = "artifact-refresh"];
```
ezimanyi: Does `keel` require kebab case, or will it also properly read snake case?

The reason I ask is that all the other services take config as snake case; this will likely get confusing if we start sharing any of these protos between services.  The other services are on Spring Boot 2 and all still are configured by snake case; I think the same should work for `keel`.  If we do want to generate output in kebab case it probably makes sense to do this for all the services at once.

ezimanyi (commented):
Thanks for doing this!  In general this looks great.  I've left a couple of larger questions, as well as a few tiny comments.  (The attention to detail on the proto comments is just because they end up in user-facing documentation.)

internal/convert/echo.go
maggieneterval: +1, and let's also consider adding documentation on enabling `keel` to the [section](https://github.com/spinnaker/kustomization-base#enable-optional-services) that documents enabling `fiat` and `kayenta`.

api/proto/config/managed_delivery.proto
```diff
@@ -0,0 +1,113 @@
...
+import "storage/sql.proto";
+
+message ManagedDelivery {
+  // Whether ManagedDelivery is enabled.
+    google.protobuf.BoolValue enabled = 1;
+    // Service discovery done through Eureka
+    Eureka eureka = 2;
+    // Keel specific configuration
+    KeelConfig keel = 3;
+
```
maggieneterval:
```suggestion
```

maggieneterval (commented):
+1 to @ezimanyi's comments, otherwise looks 💯! Thanks!

api/proto/config/managed_delivery.proto
luispollo: It should handle snake case just fine, since this is Spring Boot and there's nothing specific we're doing in keel. We were just following the best practice from the Spring docs. 🤷‍♂️

nimakaviani: great thanks for chipping in. Will make it snake case then.

nimakaviani:
Thanks very much @ezimanyi & @maggieneterval for the quick turn around on providing feedback. I have addressed most of your comments. The one outstanding item is related to `ServiceSettings` and deferring to service discovery where possible. will come back to it before the end of the week. thanks again.

internal/convert/echo.go
nimakaviani: on the above, service setting for Kayenta seems to be enabled for orca and gate if `halconfig` has `h.GetCanary().GetEnabled()` set to `true` (see [Here](https://github.com/spinnaker/kleat/blob/master/internal/convert/orca.go#L50-L59)). And frankly speaking, I like it better  than to have to do extra tweaking and setting of the configuration files in an override.

is there an ordering concern here, that makes the override a preferred choice? if not, then for `Keel` also, if we enable it similar to how we do for Kayenta and make it depend on the value of `h.GetManagedDelivery().GetEnabled()`, it will be less tweaking of the configuration files and easier for the end users. wdyat?

maggieneterval: Thanks for pointing that out @nimakaviani, I had completely forgotten where we landed on this, but based on this [PR](https://github.com/spinnaker/kleat/pull/67) from May (which feels like it was from 5 years ago instead of 5 months ago!), I think you're right -- looks like we decided to shift the `enabled` flag into any service that cares whether an optional service is enabled based on the halconfig input, but otherwise all other service discovery information should be manually overridden. Does that sound right to you, @ezimanyi?

nimakaviani: ok great, if that is the case then I will update this PR to follow the same conventions. thanks!

nimakaviani: @ezimanyi, ping on the above and also on https://github.com/spinnaker/kustomization-base/pull/32.

nimakaviani: great, I think this PR is good to go then. I will fix spinnaker/kustomization-base#32 as well.

maggieneterval: Thanks! I'm happy to merge this PR now, but we should probably hold off on merging the corresponding `kustomization-base` PR until the `keel` Docker image lives in a Spinnaker-owned repository.

maggieneterval (approved):

internal/convert/echo.go
achew22: @maggieneterval I think this is now unblocked because https://github.com/spinnaker/kustomization-base/pull/32 is submitted.

Since there has been some time since this thread was last ping'd. Are there any more things @nimakaviani would have to do before this could be merged? Thanks so much!

maggieneterval: Thanks, @achew22! Since @ncknt and @german-muzquiz are the new code owners, so I will defer to them for final approval and merge.

nimakaviani:
once tests pass, this should be good to go.

german-muzquiz (approved):

PR #146, opened 2020-10-05
feat(aws): Adding property externalId to AWS account config for assuming roles.

ivanphdz:
Enabling the use of the externalId parameter when assuming a role prevents the "confused deputy" problem.

https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user_externalid.html#external-id-purpose

related to: 
- https://github.com/spinnaker/halyard/pull/1805
- https://github.com/spinnaker/clouddriver/pull/4968


ezimanyi (approved):
Looks great, thanks!

PR #161, opened 2021-01-19
Plugin compatibility validation command

edgarulg:
This PR is about include a new command in kleat named `kleat plugins validate`.

This command will validate if the plugins configured in a halconfig has compatibility issues with the Spinnaker version in the halconfig and output any issue found.

`kleat plugins validate` uses the interfaces defined here: https://github.com/armory/plugin-tools to get compatibility metadata. The validator makes http calls to the repositories configured in the halconfig to get the compatibility metadata to evaluate.


docs/docs.md
```diff
@@ -5984,6 +5993,117 @@ Configuration for webhooks.
...
+<a name="proto.spinnaker.Repository"></a>
+
+### Repository
+Spinnaker plugin repository flags
+
+
+| Field | Type | Label | Description |
+| ----- | ---- | ----- | ----------- |
+| id | [string](#string) |  |  |
+| url | [string](#string) |  |  |
```
german-muzquiz: Description of fields is missing. It's taken from the comments in `.proto` files: https://github.com/spinnaker/kleat/blob/master/api/proto/security/trust_store.proto#L17

german-muzquiz (commented):
I think it makes sense to validate plugins compatibility also when running [`kleat generate`](https://github.com/spinnaker/kleat/blob/master/cmd/generate.go), so that users are aware that their halconfig needs to be fixed. Maybe just printing warning messages for bad plugins instead of failing service config generation?

docs/docs.md
edgarulg: I included the description for Plugins & Repositories in the `.proto`

edgarulg: I'm totally agree on include the plugin compatibility validation in the `kleat generate`, but the compatibility validation make http calls to the plugin repository and kleat is suppose to not depend on external resources. That's why I didn't include it in the `kleat generate`.

But I've been defining a mechanisms to save the compatibility metadata locally and use that information to validate compatibility, in that way we may include it in the `kleat generate`. I am planning to open another PR with those changes.

german-muzquiz (approved):

PR #175, opened 2021-05-14
Fix wrong type of `providers.aws.bakeryDefaults.baseImages[].virtualizationSettings`

yyuu:
Rosco manages `virtualizationSettings` for `aws` as a list of settings where an item contains regional VM settings, like follows.

```yaml
    aws:
      bakeryDefaults:
        baseImages:
        - baseImage:
            id: ubuntu
            shortDescription: v12.04
            detailedDescription: Ubuntu Precise Pangolin v12.04
            packageType: deb
            templateFile: aws-ebs.json
          virtualizationSettings:
          - region: us-east-1
            virtualizationType: hvm
            instanceType: t2.micro
            sourceAmi: ami-d4aed0bc
            sshUserName: ubuntu
            spotPrice: "0"
            spotPriceAutoProduct: Linux/UNIX (Amazon VPC)
```
https://github.com/spinnaker/rosco/blob/version-0.25.0/halconfig/images.yml#L4-L122

However, so far Kleat wasn't allowing to have list value at `virtualizationSettings` because the field wasn't configured for `repeated`. At very least this prevented me to configure Rosco to bake AMIs, and needed to patch generated `clouddriver.yml` and `rosco.yml` to make those services working.

It seems that `testdata` also contains wrong schema around `virtualizationSettings`. As long as I could confirm from Rosco implementation, it had `virtualizationSettings` was typed as  `List<AWSVirtuzliationSettings>`, therefore, I do believe this is a mistake made while generating testdata for `aws` provider.

```groovy
    static class AWSBakeryDefaults {
      String awsAccessKey
      String awsSecretKey
      String awsSubnetId
      String awsVpcId
      Boolean awsAssociatePublicIpAddress
      String templateFile
      BakeRequest.VmType defaultVirtualizationType
      List<AWSOperatingSystemVirtualizationSettings> baseImages = []
    }

    static class AWSOperatingSystemVirtualizationSettings {
      BakeOptions.BaseImage baseImage
      List<AWSVirtualizationSettings> virtualizationSettings = []
    }
```
https://github.com/spinnaker/rosco/blob/version-0.25.0/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/aws/config/RoscoAWSConfiguration.groovy#L65


german-muzquiz:
Thanks for pointing that out, seems that the build is failing though

yyuu:
Let me figure out the cause of failure 🙇

yyuu:
Thank you for pointing things out :bow: I have fixed the broken unit test, and now it's passing locally. I do think now it should work on CI too, however, it's not automatically running as I'm the first time contributor to this project. It's nice if someone can re-run CI job :pray:

german-muzquiz:
The main build is passing now, but `golangci` is failing: Seems that the file `rosco_test.go` needs to be formated
image
It's too bad that workflows don't run automatically to catch these earlier, I wish I had permissions to change that config :/ 

The build uses [golangci-lint](https://golangci-lint.run) for that check, in case you want to run it manually in your machine.

yyuu:
Thank you very much for guidance 🙇‍♂️ I should have checked carefully what have been done in `golangci` step 🤦  Now I ran `golangci-lint` locally and fixed the detections. It's nice if you can have another look once you get chance for it :pray:

german-muzquiz (approved):
Thanks for the fixes!

